PR #1, opened 2018-08-14
Update multiple metrics in parallel

knyar:
@fluffle, could you review this for me please?

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


app/main.go
```diff
@@ -73,7 +74,17 @@ func sync(w http.ResponseWriter, r *http.Request) {
 	}
 	defer stats.Close()
 
-	if errs := tsbridge.UpdateAllMetrics(ctx, config, sd, stats); errs != nil {
+	p, err := strconv.Atoi(os.Getenv("UPDATE_PARALLELISM"))
```
fluffle: https://play.golang.org/p/rb9vttBTHoD

This will cause problems when UPDATE_PARALLELISM is unset. I know you set it by default to 1 in the app YAML, but you probably want to also default to 1 in the code when the env var is unset…

tsbridge/metric.go
```diff
@@ -55,22 +56,41 @@ type StackdriverAdapter interface {
...
+	}
+	wg.Wait()
+	close(errchan)
+
 	for _, m := range c.Metrics() {
-		err := m.Update(ctx, sd, s)
-		if err != nil {
-			errors = append(errors, err.Error())
-		}
 		if m.Record.LastUpdate.Before(oldestWrite) {
```
fluffle: Why isn't this done in the above loop?

app/main.go
knyar: I'd rather not complicate the code by hard-coding default values. If `UPDATE_PARALLELISM` is unset, the function will return an error, which seems ok?

tsbridge/metric.go
knyar: Because `m.Record.LastUpdate` gets updated by one of those goroutines that get started behind a `SyncGroup`, so we need to wait for all of them to finish first. I've added a comment clarifying this.

PR #6, opened 2018-10-09
Integrate setup guide into README.md

myktaylor:
light editing of reset of README
added table of contents

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,191 +1,319 @@
-# Time Series Bridge (ts-bridge)
-
 Time Series Bridge is a tool that can be used to import metrics from one
-monitoring system to another. It regularly runs a specific query against
-a source monitoring system (currently only Datadog) and writes result into the
-destination system (currently only Stackdriver).
+monitoring system into another. It regularly runs a specific query against a
+source monitoring system (currently only Datadog) and writes new time series
+results into the destination system (currently only Stackdriver).
 
 ts-bridge is an App Engine Standard app written in Go.
 
-## Quick Start
+# Table of Contents
 
-* Create a GCP project that will host the Time Series Bridge.
-  * We recommend running ts-bridge in a project separate from the rest of your
-    infrastructure.
-  * If you are using ts-bridge to import metrics originating from a system
-    running on GCP, you might want to run it in a different failure domain (Cloud
-    region) from the system itself.
-* Define metrics being imported in the configuration file (`app/metrics.yaml`).
-* Install [App Engine SDK for Go](https://cloud.google.com/appengine/docs/standard/go/download)
-  (follow the "Download and install the original App Engine SDK for Go" links).
-  Add it into `$PATH`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,191 +1,323 @@
...
+1.  Clone the ts-bridge source
+    *   `go get github.com/google/ts-bridge/...`
+    *   The ts-bridge source code should appear in
+        ~/gopath/src/github.com/google/ts-bridge/
+
+## End To End Test (Dev Server)
+
+1.  Ensure that you either have **Owner** permissions for the whole Cloud
+    project, or at minimum the **Monitoring Editor** role
+1.  If using a service account to authenticate, set the
```
knyar: If this is not necessary when Cloud Shell is used, perhaps we should explicitly say so?

README.md
```diff
@@ -1,191 +1,323 @@
...
+    *   `curl http://localhost:18080/sync`
+1.  Verify that no error messages are shown. Troubleshooting guide:
+
+    | Error message | Remedy |
+    | --- | --- |
+    | ERROR: StatsCollector: rpc error: code = PermissionDenied desc = The caller does not have permission | Ensure the authenticating user has at least the "Monitoring Editor" role |
+
+1.  Configure metrics by getting your
+    [API and application keys](https://docs.datadoghq.com/api/?lang=go#overview)
+    from datadog and copying over your sli queries into `metrics.yaml`
```
knyar: I think mentioning "sli queries" here (and below) might be confusing, since ts-bridge does not really care if you're importing SLI metrics or not. I'd suggest to just say "Datadog queries" instead.

README.md
myktaylor: which part wouldn't be necessary when using the cloud shell?

README.md
myktaylor: done

README.md
knyar: I believe setting `GOOGLE_APPLICATION_CREDENTIALS` is not necessary while using Cloud Shell. In general, I would suggest keeping instructions simpler and only focusing on the Cloud Shell case.

myktaylor: Is that true?  I thought Cloud Shell used the credentials of the logged-in user by default.  Wouldn't they need to set this variable if they're using a service account?  or are you saying that we should only focus on the case where the logged-in user is deploying with her own credentials?

knyar: Yes, I am assuming the user would be using their personal credentials to deploy the app. What would be the use case that involves a service account and Cloud Shell?

myktaylor: I was thinking of the setup for automated deployment, but you're right.  that is not needed for the initial setup that this document describes.  Removing.

knyar:
Thanks! Sorry for the delay in reviewing this.

Could you please squash all commits into one and force-push to this branch so that I could merge?

myktaylor:
Squashed.  Thanks!

PR #15, opened 2018-11-22
Add support for importing cumulative metrics

knyar:
This change is quite large, but I tried to split it into two commits that I think can be reviewed separately.


tsbridge/config.go
```diff
@@ -67,9 +67,16 @@ func (c *Config) Metrics() []*Metric {
 	return c.metrics
 }
 
+// ConfigOptions is a set of global options required to initialize configuration.
+type ConfigOptions struct {
+	Filename                    string
+	DatadogMinPointAge          time.Duration
```
fluffle: These two config options have to be passed around quite a lot, and all the plumbing is starting to feel a bit ... unwieldy. I don't think the current approach will scale to many more datadog-specific options. It's not really a comment for this CL, but more a more general "I think you'll need to refactor this soon" feeling. Maybe something like [envconfig](https://github.com/kelseyhightower/envconfig) would help?

datadog/metric.go
```diff
@@ -82,34 +100,82 @@ func (m *Metric) StackdriverData(ctx context.Context, since time.Time, record *r
...
+// the `from` parameter while issuing Datadog queries, and also as the `start
+// time` field in points reported for this cumulative metric to SD.
+func (m *Metric) counterStartTime(ctx context.Context, lastPoint time.Time, rec record.MetricRecord) (time.Time, error) {
+	// Start time needs to be reset regularly, since otherwise we will be querying
+	// Datadog for a time window large enough for aggregation to kick in.
+	if time.Now().Sub(rec.GetCounterStartTime()) > m.counterResetInterval {
+		var start time.Time
+		if time.Now().Sub(lastPoint) > m.counterResetInterval {
+			// If last point was too long ago (which often means this is a new metric),
+			// set counter start time at a half of the counter reset interval.
```
fluffle: Doesn't this result in starting the counter value from a cumulative sum across half the interval, rather than 0? This seems like it might result in odd graph artifacts when you apply a rate() to the counter? (I've done nothing to prove this, it's just a vague concern.)

tsbridge/config.go
knyar: I've been thinking about this as well. I agree that envconfig, or maybe just moving these into a `global` section of the .yaml config file, would be better.

datadog/metric.go
knyar: I am not sure I understand the question, so let me try to explain what happens here in more detail.

Typically when we reset start time, it's set to `lastPoint`+1s. This ensures continuity of data, and this is the most common case.

However, when `lastPoint` is too old (more than `resetInterval` old), we cannot use it as a basis for new start time, since it will make new start time still older than `resetInterval`, and it will immediately need to be moved forward again. This only happens when a new metric is added, or when writes to Stackdriver have been failing for more than `resetInterval`.

What we need to do in this rare case is to choose an arbitrary point in the recent past as the new start time. In this situation there will be a gap in data between `lastPoint` and this new start time, however metric values will be reported correctly: as usual, Datadog will return a cumulative sum of values starting from 0 at this newly selected start time.

I see three options for choosing this "arbitrary point in the recent past":
- go back the full `resetInterval`. This will ensure we backfill as much data as possible, however will also mean start time will need to be immediately reset during the next update operation (and if next update operation fails, `lastPoint` will be too old again and we'll have a short gap in data).
- just set new start time at current timestamp. As the result, we backfill no data at all, and just start a new counter from 0.
- somewhere in between: long enough to backfill some data, but not so long that a single failed write will result in a gap. This is the option I chose, and I use the mid-point between "now" and "resetInterval ago".

Does this make sense? I will restructure code a bit now, making sure that the "common case" goes first in the if/else block, and will copy some parts of this explanation into code comments.

PR #16, opened 2019-03-17
Clarify requirements for local testing

gplasky:
Without setting `SD_PROJECT_FOR_INTERNAL_METRICS`, local testing will fail with the following error when /sync is called:

```
2019/03/16 23:23:29 ERROR: Please define SD_PROJECT_FOR_INTERNAL_METRICS if not running on App Engine
INFO     2019-03-16 23:23:29,119 module.py:861] default: "GET /sync HTTP/1.1" 500 75
```


knyar:
Thank you!

PR #23, opened 2020-02-06
Introduce support for InfluxDB gauge metrics

yanske1:
Add initial support for InfluxDB gauge metrics. This code currently lives in isolation and is not hooked up to the existing config parsing logic. Given a gauge metric metric, we can now build an Influx query and convert the time series points into Stackdriver points.

What this PR does not contain:
- Config parsing for Influx metrics
- Influx metric documentation
- Support for time aggregated metrics in Influx
- Support for cumulative metrics

Design doc: go/ts-bridge-influx-design-doc


influxdb/metric.go
```diff
@@ -0,0 +1,224 @@
...
+	// by applying an offset. We'll need to apply filtering once cumulative
+	// metrics come into play.
+	return m.metricDescriptor(), m.convertTimeSeries(points), nil
+}
+
+// buildQuery creates a InfluxDB query from the metric query definition,
+// wrapped in the given time-interval, inclusive of start time.
+func (m *Metric) buildQuery(startTime, endTime time.Time) client.Query {
+	query := fmt.Sprintf(
+		"SELECT * FROM (%s) WHERE time >= %s AND time < %s",
```
knyar: I think you should be able to use `%d` rather than `%s` for integer parameters here, instead of converting them to strings using `FormatInt`.

influxdb/metric.go
```diff
@@ -0,0 +1,224 @@
...
+	return points, nil
+}
+
+// convertPoint converts a parsed InfluxDB point into a Stackdriver point.
+func (p *point) convertPoint() *monitoringpb.Point {
+	// For gauge metrics without time aggregations, we can treat the timestamps
+	// given by Influx as EndTime for the Stackdriver point.
+	return &monitoringpb.Point{
+		Interval: &monitoringpb.TimeInterval{
+			EndTime: p.protoTimestamp(),
```
knyar: I think `timestamp.TimestampProto()` is the recommended way to convert `time.Time` to a `Timestamp` proto.

influxdb/metric.go
yanske1: whoops good catch, done

influxdb/metric.go
yanske1: done

knyar:
LGTM. I am happy to merge this, but would like to give @myktaylor a chance to comment on this first.

influxdb/metric.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+type point struct {
+	timestamp time.Time
+	value     float64
+}
+
+// parseSeriePoints parses points from an InfluxDB series into a slice of
+// timestamp-value pairs.
+func parseSeriePoints(serie models.Row) ([]point, error) {
```
myktaylor: Is this domain-specific lingo? Why are all these named "serie" instead of "series"?

yanske1: O I thought the singular form of "series" was "serie". Updated.

PR #24, opened 2020-02-07
Support time aggregated queries for InfluxDB

yanske1:
When gauge metrics are aggregated by time in InfluxQL (e.g. `SELECT * FROM foo GROUP BY time(10s)`, we want to set the end time of a time series point to be the end time of the time aggregated interval, and filter out intervals that are not yet ready to be imported.

This is step also lays down the ground work for cumulative metrics, as they will also be time aggregated.


influxdb/metric.go
```diff
@@ -147,10 +145,46 @@ func (m *Metric) metricDescriptor() *metricpb.MetricDescriptor {
...
+	var filteredPoints []point
+	for _, p := range points {
+		// Once cumulative metrics come in play, we'll filter using lastPoint.
+		// For now we can rely on our constructed query to only fetch the
+		// latest points.
+
+		// For time-aggregated InfluxQL queries with interval i, Influx returns
+		// points with timestamp x, denoting the aggregated value from time
+		// [i, i + x).
+		intervalEnd := p.timestamp.Add(queryInterval)
```
knyar: Instead of throwing away `intervalEnd`, and later separately parsing and adding `queryInterval` to each point (in `convertPoint`), should we just overwrite `p.timestamp` here?

influxdb/config.go
```diff
@@ -0,0 +1,80 @@
...
+	matches := re.FindAllStringSubmatch(c.Query, -1)
+
+	if len(matches) != 1 {
+		return 0, fmt.Errorf("query %s has %d time groupings, expected 1", c.Query, len(matches))
+	}
+	if len(matches[0]) != 3 {
+		return 0, fmt.Errorf("query %s expected {match} {magnitude} {unit} match, got %v", c.Query, matches[0])
+	}
+
+	return parseDuration(matches[0][1], matches[0][2])
```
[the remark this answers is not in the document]
yanske1: Good catch, done

influxdb/metric.go
yanske1: Good idea, added.

knyar:
Thanks!

PR #25, opened 2020-02-11
Support for cumulative InfluxDB metrics

yanske1:
Add support for cumulative metrics with or without time aggregation. The main logic update here is passing down the start time from Datastore when converting Influx points to SD points, and resetting the counter start time periodically.


influxdb/metric.go
```diff
@@ -138,82 +185,101 @@ func (m *Metric) buildQuery(startTime, endTime time.Time) client.Query {
...
+			if err != nil {
+				return nil, err
+			}
+
+			p.timestamp = p.timestamp.Add(queryInterval)
+		}
+
+		// If this (time-aggregated) time interval hasn't finished accumulating
+		// data, wait for it to complete next time around.
+		if p.timestamp.UnixNano() > endTime.UnixNano() {
```
knyar: Is this necessary for cumulative metrics that are not time-aggregated? If not, perhaps this should be moved to the `if m.config.TimeAggregated` block above?

influxdb/metric.go
```diff
@@ -138,82 +185,101 @@ func (m *Metric) buildQuery(startTime, endTime time.Time) client.Query {
...
 	et, err := ptypes.TimestampProto(p.timestamp)
 	if err != nil {
 		return nil, err
 	}
 
+	interval := &monitoringpb.TimeInterval{
+		EndTime: et,
+	}
+	if m.config.Cumulative {
+		st, err := ptypes.TimestampProto(startTime)
```
knyar: I think you can just assign to `interval.StartTime` here instead of creating a temporary variable.

influxdb/metric.go
yanske1: Moved!

influxdb/metric.go
yanske1: Done, thanks

knyar:
Thanks!

PR #26, opened 2020-02-11
Integrate InfluxDB configs into ts-bridge

yanske1:
Adds InfluxDB config to the list of configs we parse for in ts-bridge, and consolidates some of the logic used to merge a specific SourceMetric (e.g. Datadog, Influx) into the list of overall config metrics.

Also used the Datadog environment variables for InfluxDB metrics. Though these variables may be unique to Datadog & Influx (especially COUNTER_RESET_INTERVAL), it might make sense to consolidate them for now considering it's unlikely for a customer to have both Datadog and Influx metrics, and want different options for them.

This will be followed up by a documentation update.


influxdb/config.go
```diff
@@ -24,12 +24,15 @@ import (
...
-	Query          string
-	Database       string
-	Endpoint       string
-	Username       string
-	Password       string
-	TimeAggregated bool
+	Query    string `validate:"nonzero"`
+	Database string `validate:"nonzero"`
+	Endpoint string `validate:"nonzero"`
+	Auth     struct {
```
knyar: Any reason for moving username & password into a separate substruct?

yanske1: Mainly to make it more obvious that they're a related concept -- if you don't want auth you can just not have the whole auth block when you're copying over the YAML example, instead of mentally having to ignore both the username and password. This might be more helpful when other forms are authentication (e.g. token) are introduced.

No strong opinions though, what do you think?

knyar: I agree that it would have been useful if we had several authentication types, in which case we could have `BasicAuth` and `TokenAuth` substructs, with a check that only one of them is defined for a given metric. 

But since now we only have username & password, it feels like an unnecessary abstraction. I would probably suggest to keep everything on the same level for now.

yanske1: Makes sense, thanks!

knyar:
Oops, sorry for merging this (and #25) without waiting for an LGTM from @myktaylor.

Myk, I think you can still comment and we'll ask @yanske1 to prepare another PR addressing your comments. What do you think?

PR #27, opened 2020-02-12
Update documentation with InfluxDB metrics

yanske1:
Add documentation for InfluxDB as a metric source, and share environment variables previously only used for Datadog.

This PR splits the documentation for source metrics into its own README since I felt like the root README was getting too long and complex with the new Influx documentations. Hopefully this is more understandable, but if not I can merge it back into the root README. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 Time Series Bridge is a tool that can be used to import metrics from one
 monitoring system into another. It regularly runs a specific query against a
-source monitoring system (currently only Datadog & InfluxDB) and writes
+source monitoring system (currently Datadog & InfluxDB) and writes
 new time series results into the destination system (currently only
 Stackdriver).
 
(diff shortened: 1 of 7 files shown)


influxdb/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+For metrics that have measurements more often than every minute, you might
+consider time aggregating the query to avoid importing too many points. For
+example, a gauge metric query such as:
+
+`SELECT free_disk_space FROM sys_disk`
+
+can be time aggregated as the following:
+
+`SELECT MEAN(free_disk_space) FROM sys_disk GROUP BY time(1m)`.
```
knyar: Perhaps mention again here that this also requires `time_aggregated` to be set to true?

README.md
```diff
@@ -262,17 +200,15 @@ the `env_variables` section of `app/app.yaml`.
...
-    to avoid [aggregation](https://docs.datadoghq.com/graphing/faq/what-is-the-granularity-of-my-graphs-am-i-seeing-raw-data-or-aggregates-on-my-graph/)
-    on Datadog side. This parameter defines how often a new start time is
-    chosen. 30 minutes should be sufficient for metrics that have a point
-    every 10 seconds. See [Cumulative metrics](#cumulative-metrics) section
-    below for more details.
+    (default is 1.5 minutes) are ignored, since the metric source might return
+    incomplete data for them if some input data is delayed.
+*   `COUNTER_RESET_INTERVAL`: while importing counters, ts-bridge needs
+    to reset 'start time' regularly to keep the query time window small enough.
+    This parameter defines how often a new start time is chosen, and is
```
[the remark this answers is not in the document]
yanske1: Done

influxdb/README.md
yanske1: Good point, added! :)

yanske1:
I'll update this diff again once the config changes are in / finalized. https://github.com/google/ts-bridge/pull/26

knyar:
LGTM, but this time I will actually wait for @myktaylor before merging.

influxdb/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+When using queries with time aggregation, ts-bridge requires the
+`time_aggregated` flag to be set to `true` in the metric's YAML configuration.
+Otherwise, unexpected errors may arise when attempting to process these time
```
myktaylor: could we be more specific about this? what errors occur and why?

influxdb/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+Currently only InfluxDB 1.x series is supported.
+
+## Cumulative metrics
+
+Cumulative metrics are supported through InfluxDB's `CUMULATIVE_SUM` function,
+which returns a montonically increasing time series with a sum of all
+measurements since the given 'start time'.
+
+This can be combined with aggregation to create counter like metrics. For
```
myktaylor: "counter like" -> "counter-like"

influxdb/README.md
```diff
@@ -0,0 +1,80 @@
...
+example, a popular metric required for SLIs may be:
+
+`SELECT CUMULATIVE_SUM(COUNT(requests)) FROM nginx_access_log GROUP BY time(1m)`.
+
+Note that the above style of nested aggregation (i.e. having a function inside
+of `CUMULATIVE_SUM()`) is only supported in InfluxDB 1.7. To achieve the same
```
myktaylor: 1.7 -> 1.7 and newer

influxdb/README.md
```diff
@@ -0,0 +1,80 @@
...
+effect in InfluxDB 1.6, a subquery can be defined instead:
+
+`SELECT CUMULATIVE_SUM(*) FROM (SELECT COUNT(request) FROM nginx_access_log
+GROUP BY time(1m))`.
+
+When using the above example, make sure to set `cumulative` to `true`, and
+`time_aggregated` to `true` if applicable.
```
myktaylor: remove "if applicable"? if they're using the above example, don't we need to set the two properties to true?

influxdb/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Zero value intervals
+
+For time aggregated cumulative queries, if no rows exist within the queried
+time frame, then the cumulative query will return no results. Often however,
+this is not an indication of having no data, but rather having timestamps with
```
myktaylor: rather having -> rather of having

influxdb/README.md
```diff
@@ -0,0 +1,80 @@
...
+value 0. To avoid this, we can use the
```
myktaylor: why would a timestamp be 0?

yanske1: Ah in this case the timestamp would be whatever the interval is aligned to, but it should have the corresponding value of 0. Updated to make it more clear

influxdb/README.md
yanske1: InfluxDB only goes up to version 1.7 (then it jumps to 2.0, which we don't support)

datadog/README.md
```diff
@@ -0,0 +1,84 @@
...
+    ts-bridge, some import operations will fetch 0 new points. For example, if
+    your query is producing a 10-minute ratio ( `xxx.rollup(sum, 600) /
+    yyy.rollup(sum, 600)`) and you are using the default importing period (1
+    minute), ts-bridge will still issue the query every minute, however Datadog
+    will only return a single point once every 10 minutes.
+*   If you are not using the `rollup` function, Datadog will return points at
+    maximum possible resolution (unless the query covers a very long time
+    interval). Please keep in mind that Datadog might return more than 1 point
+    per minute, and all points will be written to Stackdriver, even though
+    Stackdriver does not allow querying with
```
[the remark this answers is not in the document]
yanske1: I think so. This is a copy-paste of the previous README, but just ported to `datadog/README.md`

app/app.yaml
```diff
@@ -15,13 +15,13 @@ env_variables:
   SD_PROJECT_FOR_INTERNAL_METRICS: ""
   # Number of metrics to update in parallel. Must be between 1 and 100 (chosen arbitrarily).
   UPDATE_PARALLELISM: 1
-  # Points received from Datadog that are too fresh will be discarded to allow data to be fully processed and aggregated
-  # on Datadog side. This variable defines the threshold for "too fresh".
-  DATADOG_MIN_POINT_AGE: "2m"
```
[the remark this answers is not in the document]
yanske1: As long as those users are not overriding or maintaining the environment variables using something that is not the app YAML, this should sync over fine.

myktaylor:
ah, because if they have modified the value in app.yaml, it will generate a
merge conflict and require manual attention?

yanske1:
yeah (and even if it doesn't, it is only a name change in this case and won't affect any metric importing)

myktaylor:
name changes can be problematic. they can subtly break user code. in this
case, though, the git merge logic should highlight the change so there
won't be any invisible breakage.

PR #28, opened 2020-04-16
Document Influx VM setup

yanske1:


influxdb/README.md
```diff
@@ -78,3 +78,16 @@ provided such as:
...
 FILL(0)`.
+
+## GCP Setup
+
+A common InfluxDB setup will involve having InfluxDB running in a GCE VM, if
+the user is not using InfluxDB cloud. To connect ts-bridge, an App Engine
+application to a GCE VM, a
+[serverless VPC](https://cloud.google.com/vpc/docs/configure-serverless-vpc-access)
+can be configured for App Engine, and ts-bridge can then be pointed to the VM's
+internal IP address.
```
knyar, line 89: I would suggest to explicitly mention that this is only necessary if InfluxDB port needs to be unreachable from public internet.

knyar:
Did you intend to add changes to `.travis.yml` into this PR as well?

yanske1:
Yeah, I'm suspecting mockgen is not being held to a revision is resulting in go generate ./.. changes, WIP at the moment

influxdb/README.md
yanske1: Good idea, added

yanske1:
Added a TODO for now for the travis

Issue #33, opened 2020-04-29
Unsupported google.golang.org/appengine version


Dnefedkin:
Current version of google.golang.org/appengine  leads to "not an App Engine context" errors for all Cloud Datastore calls that ts-bridge is doing

PR #34, opened 2020-04-29
Updating google.golang.org/appengine dependency.

Dnefedkin:
Updating google.golang.org/appengine dependency., as the current version leads to "not an App Engine context" error,  it was discovered while setting us ts-bridge in one of the client environments. See also #33


Dnefedkin:
All the changes in this pull request have been generated by executing "dep ensure -update google.golang.org/appengine"

Dnefedkin:
This fix has been tested in a customer environment and the customer has proved that this has helped to fix the issue.

PR #36, opened 2020-08-04
Refactor record.go to allow swappable storage backends

Temikus:
This is the first step towards adding BoltDB as a supported datastore.

- Refactor all the datastore logic into a separate file
- Create a set of abstract storage functions that can return different pluggable backends
- Add `STORAGE_ENGINE` config option that defaults to GAE datastore

This is an early stab at it, but since I'm relatively new to Golang I wanted to get feedback early rather than later.
@Dnefedkin can you take a look and let me know what you think?

Thank you.


record/record.go
```diff
@@ -35,106 +38,30 @@ type MetricRecord interface {
...
-
-// UpdateSuccess updates metric status in Datastore with a given message.
-func (m *DatastoreMetricRecord) UpdateSuccess(ctx context.Context, points int, msg string) error {
-	log.Infof(ctx, "%s: %s", m.Name, msg)
-	m.LastStatus = fmt.Sprintf("OK: %s", msg)
-	m.LastAttempt = time.Now()
-	if points > 0 {
-		m.LastUpdate = time.Now()
+// NewStoredMetricRecord provides an abstraction for pluggable storage engines implementing MetricRecord interface
+func NewStoredMetricRecord(ctx context.Context, name, query string, storage string) (MetricRecord, error){
```
knyar, line 42: Do you anticipate the need to use several different storage engines as part of the same ts-bridge deployment? If not, these wrappers seem unnecessary and a cleaner way might be to define a storage provider interface (it seems like it only needs to have two functions: `NewRecord` and `CleanupRecords`), make two implementations of it (one for Datastore, the other for BoltDB), and instantiate one of the two providers when the app is initialized.

tsbridge/config.go
```diff
@@ -98,6 +102,8 @@ func NewConfig(ctx context.Context, opts *ConfigOptions) (*Config, error) {
 		return nil, fmt.Errorf("configuration file validation error: %s", err)
 	}
 
+	c.Storage = opts.StorageEngine
```
knyar, line 105: Having a `storage` parameter in the configuration file that is always overwritten with the `STORAGE_ENGINE` environment variable is pretty confusing.

I'd suggest to choose one configuration mechanism (file or environment variable) and stick with it. If the long term plan is to allow ts-bridge to be deployed outside of App Engine, perhaps it's going to be simpler to move all options into the configuration file and get rid of environment variables completely.

record/record.go
Temikus: Gotcha. Abstracted the storage into a `StorageManager`, PTAL.

tsbridge/config.go
Temikus: Agreed it is confusing. I'll work with the env variable only for now (removed it from the parsed struct) but I do plan to address the ENV vars wholesale at a later point. PTAL at the new implementation.

Temikus:
@knyar Can you take a second look?

Also - with the current implementation I'm running into mockgen issues on `metric.go`:
```
λ go generate tsbridge/metric.go                                                        (1)
Reflection: can't yet turn interface { GetCounterStartTime() time.Time; GetLastUpdate() time.Time; SetCounterStartTime(context.Context, time.Time) error; UpdateError(context.Context, error) error; UpdateSuccess(context.Context, int, string) error } (interface) into a model.Type
Reflection: can't yet turn interface { GetCounterStartTime() time.Time; GetLastUpdate() time.Time; SetCounterStartTime(context.Context, time.Time) error; UpdateError(context.Context, error) error; UpdateSuccess(context.Context, int, string) error } (interface) into a model.Type
Reflection: can't yet turn interface { GetCounterStartTime() time.Time; GetLastUpdate() time.Time; SetCounterStartTime(context.Context, time.Time) error; UpdateError(context.Context, error) error; UpdateSuccess(context.Context, int, string) error } (interface) into a model.Type
2020/08/06 15:17:50 Loading input failed: exit status 1
tsbridge/metric.go:43: running "mockgen": exit status 1
```
Interestingly enough though, `mockgen` doesn't display any errors and generates a mock:
```
λ mockgen -source=tsbridge/metric.go > /dev/null && echo $?
0
```

Presumably it's due to https://github.com/golang/mock/issues/153 and the alias I used for the interface, but I'm not entirely sure how to address it. Any pointers would be appreciated if you ran into this before.

Thanks!

app/main.go
```diff
@@ -34,6 +35,9 @@ import (
 	"google.golang.org/appengine/log"
 )
 
+// Global Storage Manager object to be used while running the app
+var StorageManager storage.Manager
```
dparrish, line 39: This should be `storageManager` as it isn't (and can't) be exported from main.

datastore/datastore_record.go
```diff
@@ -12,22 +12,20 @@
...
+	"time"
 )
 
-//go:generate mockgen -destination=../mocks/mock_metric_record.go -package=mocks github.com/google/ts-bridge/record MetricRecord
+//go:generate mockgen -destination=../mocks/mock_metric_record.go -package=mocks github.com/google/ts-bridge/datastore MetricRecord
 
-// MetricRecord is an interface implemented by DatastoreMetricRecord.
-type MetricRecord interface {
+// MetricRecord is an interface implemented by StoredMetricRecord.
+type MetricRecord = interface {
```
dparrish, line 28: Remove the `=`

dparrish: Also I think the MetricRecord interface should be defined outside the datastore specific code, perhaps `record/record.go`, because it will be implemented by the Datastore code as well as future backends.

storage/storage.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// StorageManager interface implemented by associated Manager
+type Manager interface {
+	NewMetricRecord(ctx context.Context, name, query string) (MetricRecord, error)
+	CleanupRecords(ctx context.Context, valid []string) error
+}
+
+// MetricRecord is an interface implemented by StoredMetricRecord.
+type MetricRecord = interface {
```
dparrish, line 31: Remove `=`

dparrish: Is this supposed to be different to the MetricRecord interface defined by datastore?

app/main.go
Temikus: Acknowledged. Fixed.

datastore/datastore_record.go
Temikus: @dparrish So this use of Go 1.9 alias is intentional, since otherwise I get:

```
# github.com/google/ts-bridge/app
app/main.go:166:18: cannot use "github.com/google/ts-bridge/datastore".New() (type *"github.com/google/ts-bridge/datastore".Manager) as type storage.Manager in assignment:
	*"github.com/google/ts-bridge/datastore".Manager does not implement storage.Manager (wrong type for NewMetricRecord method)
		have NewMetricRecord("context".Context, string, string) ("github.com/google/ts-bridge/datastore".MetricRecord, error)
		want NewMetricRecord("context".Context, string, string) (storage.MetricRecord, error)
app/main.go:169:18: cannot use "github.com/google/ts-bridge/datastore".New() (type *"github.com/google/ts-bridge/datastore".Manager) as type storage.Manager in assignment:
	*"github.com/google/ts-bridge/datastore".Manager does not implement storage.Manager (wrong type for NewMetricRecord method)
		have NewMetricRecord("context".Context, string, string) ("github.com/google/ts-bridge/datastore".MetricRecord, error)
		want NewMetricRecord("context".Context, string, string) (storage.MetricRecord, error)
FAIL	github.com/google/ts-bridge/tsbridge [build failed]

Process finished with exit code 2

# github.com/google/ts-bridge/tsbridge [github.com/google/ts-bridge/tsbridge.test]
tsbridge/config_test.go:34:81: cannot use "github.com/google/ts-bridge/datastore".New() (type *"github.com/google/ts-bridge/datastore".Manager) as type storage.Manager in field value:
	*"github.com/google/ts-bridge/datastore".Manager does not implement storage.Manager (wrong type for NewMetricRecord method)
		have NewMetricRecord("context".Context, string, string) ("github.com/google/ts-bridge/datastore".MetricRecord, error)
		want NewMetricRecord("context".Context, string, string) (storage.MetricRecord, error)
tsbridge/config_test.go:73:97: cannot use "github.com/google/ts-bridge/datastore".New() (type *"github.com/google/ts-bridge/datastore".Manager) as type storage.Manager in field value:
	*"github.com/google/ts-bridge/datastore".Manager does not implement storage.Manager (wrong type for NewMetricRecord method)
		have NewMetricRecord("context".Context, string, string) ("github.com/google/ts-bridge/datastore".MetricRecord, error)
		want NewMetricRecord("context".Context, string, string) (storage.MetricRecord, error)
tsbridge/metric_test.go:123:86: cannot use "github.com/google/ts-bridge/datastore".New() (type *"github.com/google/ts-bridge/datastore".Manager) as type storage.Manager in argument to NewMetric:
	*"github.com/google/ts-bridge/datastore".Manager does not implement storage.Manager (wrong type for NewMetricRecord method)
		have NewMetricRecord("context".Context, string, string) ("github.com/google/ts-bridge/datastore".MetricRecord, error)
		want NewMetricRecord("context".Context, string, string) (storage.MetricRecord, error)
tsbridge/metric_test.go:163:84: cannot use "github.com/google/ts-bridge/datastore".New() (type *"github.com/google/ts-bridge/datastore".Manager) as type storage.Manager in argument to NewMetric:
	*"github.com/google/ts-bridge/datastore".Manager does not implement storage.Manager (wrong type for NewMetricRecord method)
		have NewMetricRecord("context".Context, string, string) ("github.com/google/ts-bridge/datastore".MetricRecord, error)
		want NewMetricRecord("context".Context, string, string) (storage.MetricRecord, error)
```

Perhaps I'm missing something here though?

storage/storage.go
Temikus: No, it is not supposed to be different. I thought interface should be implemented both on the producer and consumer side. I think I'm missing something here - can you elaborate?

datastore/datastore_record.go
dparrish: Yep, you're hiding incomplete code. datastore.Manager (the struct) does not have all the methods necessary to implement the storage.Manager interface, so the compiler says you're doing it wrong.

You need to add the missing methods to datastore.Manager

storage/storage.go
dparrish: An interface in Go defines what you *can* or *will* call, not what's available. 

A struct that just has some methods doesn't need to define an interface to say what it does, that's part of the struct itself. Instead you define an interface at the consumer side to say what you're going to call and any struct that has those methods can be used.

Interfaces are also used to abstract away implementation details. For example the io.Writer interface https://godoc.org/io#Writer is defined so that any object that implements that exact Write method can be passed to the various os functions, and the os level doesn't have to care what the object does with that data.

datastore/datastore_record.go
Temikus: Ah, I see - fixed up the `StoredMetricRecord` and removed the interface. Thanks!

storage/storage.go
Temikus: Gotcha, this is super-helpful, removed the redundant interface.

app/main.go
```diff
@@ -18,14 +18,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/google/ts-bridge/storage"
```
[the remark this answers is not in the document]
Temikus: Done

app/main.go
```diff
@@ -34,6 +35,9 @@ import (
 	"google.golang.org/appengine/log"
 )
 
+// Global Storage Manager object to be used while running the app
+var storageManager storage.Manager
```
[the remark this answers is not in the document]
Temikus: @knyar So the problem here is that Storage is initialised in the ConfigOptions, not the Config itself:

https://github.com/google/ts-bridge/blob/fe6db8d46812db782dbdb73784943dd11fd983f8/app/main.go#L173-L179

And the config returned in `tsbridge` module is parsed from YAML:
https://github.com/google/ts-bridge/blob/fe6db8d46812db782dbdb73784943dd11fd983f8/tsbridge/config.go#L95

As a result, if I add something to the `Config` struct, it may cause parsing override issues (as you've correctly pointed to me before). As such, I added one package-level var in `main.go` to refer to storage engine.

An alternative approach could be:
- Creating a new config map and method to load it (e.g. `NewRuntimeConfig` and `NewMetricConfig`)
- Leaving it as-is and adressing more thoughtfully when we address the env variables vs. YAML config

But again, I'm new to Go - so if I'm missing anything or you have another approach in mind - just let me know! :)

app/main.go
```diff
@@ -156,10 +160,22 @@ func newConfig(ctx context.Context) (*tsbridge.Config, error) {
 		return nil, fmt.Errorf("Could not parse COUNTER_RESET_INTERVAL: %v", err)
 	}
 
+	storageEngine := os.Getenv("STORAGE_ENGINE")
+	switch storageEngine {
+	case "datastore":
+		storageManager = datastore.New()
+	case "":
```
Dnefedkin, line 167: As nowhere in the default configs we explicitly assign STORAGE_ENGINE,  os.Getenv("STORAGE_ENGINE") will return an empty string and this will lead to "defaulting to GAE datastore".  

This is fine, but let's make sure we explicitly assign STORAGE_ENGINE=datastore in app.yaml and maybe also reference to it in README.md #global-settings , so that potential customers of ts-bridge are aware, that there is a setting like that, even though it's currently allowing just 1 option to choose from.

Dnefedkin:
Added a comment regarding documenting STORAGE_ENGINE , otherwise LGTM.

app/main.go
Temikus: Done.

app/main.go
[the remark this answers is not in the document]
Temikus: Agreed, rolled it into a helper function.

Temikus:
Fixed up the mocks by running the same version as CI. Merging.

PR #37, opened 2020-08-30
Update the app for GAE Go 1.12+ runtime

Temikus:
Since `bbolt` which we plan to use as a storage engine requires Go 1.13+, we need to update the app to support GAE 1.12+ runtime (which we would've inevitably needed to do anyway).

Rough outline of changes here:
- Moving to go modules
- Changing the runtime to 1.13
- Migrate logging to logrus
- Remove urlfetch, appengine.Main() and other GAE library-specific routines
- Get runtime info from ENV instead of using `appengine` lib
- Change some config paths and associated docs due to different PWD

NOTE TO REVIEWER:
I know this is a large PR, but the migration scope is pretty large so I want to avoid merging commits that will make the project unbuildable. As such, I've partitioned the PR into small commits that should be limited in scope and easier to review. I plan to squash them after the review is finished.


app/main.go
```diff
@@ -32,7 +32,7 @@ import (
 
 	"github.com/dustin/go-humanize"
 	"google.golang.org/appengine"
-	"google.golang.org/appengine/log"
+	log "github.com/sirupsen/logrus"
```
Dnefedkin, line 34: Is there a specific reason why logrus was choosen?

E.g. https://github.com/go-logr/logr is an abstract interface that supports multiple implementations including logrus.

.travis.yml
```diff
@@ -1,19 +1,17 @@
...
-  - curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-257.0.0-linux-x86_64.tar.gz | tar xfz - -C $HOME
-  - $HOME/google-cloud-sdk/bin/gcloud components install app-engine-python app-engine-go cloud-datastore-emulator
-  # TODO: lock a version of mockgen.
-  - go get github.com/golang/mock/mockgen
-  - curl -L -s https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 -o $GOPATH/bin/dep
-  - chmod +x $GOPATH/bin/dep
+  - curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-307.0.0-linux-x86_64.tar.gz | tar xfz - -C $HOME
+  - $HOME/google-cloud-sdk/bin/gcloud components install beta cloud-datastore-emulator
+  # JRE is needed for the datastore emulator
+  - sudo apt-get install openjdk-8-jre-headless
```
Dnefedkin, line 10: Emulator is not starting in travis build with "Unable to prepare the data directory for the emualtor" error. I think you saw https://github.com/GoogleCloudPlatform/cloud-sdk-docker/issues/144 and  decided to install JRE 8 , right?


Add   - java -version    after the installation, maybe it will be pointing to 1.7 even after install, and you need to do something like sudo update alternatives  to explicitly specify 1.8

datastore/emulator.go
```diff
@@ -0,0 +1,94 @@
...
+	"net/http"
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+// Datastore emulator for use in tests
+//
+// Background: this was devised due to aetest package no longer being available for 1.12+ runtimes as dev_appserver.py
+// cannot run runtimes past go111
```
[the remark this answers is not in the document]
Temikus: Yep, this was raised before multiple times:
https://groups.google.com/g/google-appengine-go/c/kHxZ9zIb_QE/m/QjI0lnorBQAJ
https://github.com/googleapis/google-cloud-go/issues/644

I'll reference the issue in the code as well.

app/main.go
```diff
@@ -116,10 +139,11 @@ func cleanup(w http.ResponseWriter, r *http.Request) {
 		metricNames = append(metricNames, m.Name)
 	}
 
-	storageManager, err := loadStorageEngine()
+	storageManager, err := loadStorageEngine(ctx)
```
knyar, line 142: Do you need a second copy of the storage engine here?

app/main.go
```diff
@@ -172,27 +203,37 @@ func newConfig(ctx context.Context) (*tsbridge.Config, error) {
...
+		return datastore.New(ctx), nil
 	default:
 		return nil, fmt.Errorf("unknown storage engine selected: %s", storageEngine)
 	}
 }
+
+// Check if we're running in AppEngine by checking GAE_ENV variable
+func isAppEngine() bool {
+	_, set := os.LookupEnv("GAE_ENV")
+	if set {
```
knyar, line 234: Perhaps you can just return `set`?

datadog/metric_test.go
```diff
@@ -33,23 +33,17 @@ import (
...
 func TestMain(m *testing.M) {
-	var done func()
-	var err error
-	testCtx, done, err = aetest.NewContext()
-	if err != nil {
-		panic(err)
-	}
-
+	DSStop, _ := datastore.Emulator(8081)
+	testCtx = context.Background()
```
knyar, line 44: I don't think you need a shared test context anymore, so perhaps each test should just call `context.Background()` independently.

datastore/emulator.go
```diff
@@ -0,0 +1,95 @@
...
+//
+//   Usage:
+//
+//   DSStop, DSReset := Emulator(8081)
+//   defer DSStop()
+//
+func Emulator(port int) (stop func(), reset func()) {
```
knyar, line 37: Do we need to start/stop this from Go tests, and re-start it for every test? If not, perhaps it would be simpler to started the emulator externally, before running `go test` (using a Makefile or a bash script)

knyar (changes requested):

app/main.go
Dnefedkin: Sounds good, let's use logrus directly at this stage.

datadog/metric_test.go
Temikus: Fair point, rewrote to use separate contexts.

datastore/emulator.go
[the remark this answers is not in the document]
Temikus: Put a reference comment in the code.

.travis.yml
Temikus: No, this was a side effect of another issue, which is now fixed (process was hanging around stuck).

datastore/emulator.go
```diff
@@ -0,0 +1,115 @@
...
+//   Usage:
+//
+//   testCtx, testCtxCancel := context.WithCancel(context.Background())
+//   err := datastore.Emulator(testCtx)
+//   testCtxCancel()
+//
+func Emulator(ctx context.Context) error {
+
+	// Set log formatter to process newlines since emulator gives a lot of output
+	log.SetFormatter(&log.TextFormatter{DisableQuote: true})
```
dparrish, line 50: This changes the global logging state, probably not what you want. logrus can create child loggers with different formats which might be good here.

or move this SetFormatter line to main().

dparrish (changes requested):

datastore/datastore_manager.go
```diff
@@ -15,25 +15,30 @@
...
 	"github.com/google/ts-bridge/storage"
-	"google.golang.org/appengine/datastore"
 	log "github.com/sirupsen/logrus"
 )
 
 // New initializes the Manager struct implementing a generic storage.Manager interface
-func New() *Manager {
-	return &Manager{}
+func New(ctx context.Context) *Manager {
+	dsClient, err := datastore.NewClient(ctx, "graphite-sandbox")
```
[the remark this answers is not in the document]
Temikus: Fixed, that's my bad.

datadog/metric_test.go
```diff
@@ -26,30 +26,25 @@ import (
...
 )
 
-var testCtx context.Context
-
 func TestMain(m *testing.M) {
-	var done func()
-	var err error
-	testCtx, done, err = aetest.NewContext()
+	testCtx, testCtxCancel := context.WithCancel(context.Background())
+	err := datastore.Emulator(testCtx)
```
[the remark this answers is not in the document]
Temikus: This is no longer relevant, though I'll keep it in mind for the future, ta!

datadog/metric_test.go
```diff
@@ -26,30 +26,25 @@ import (
...
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
 )
 
-var testCtx context.Context
-
 func TestMain(m *testing.M) {
-	var done func()
-	var err error
-	testCtx, done, err = aetest.NewContext()
+	testCtx, testCtxCancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
Temikus: done

datadog/metric_test.go
```diff
@@ -92,27 +87,31 @@ func makeTestServer(filename string) (*fixtureHandler, *httptest.Server) {
 }
 
 func TestStackdriverDataErrors(t *testing.T) {
+	testCtx := context.Background()
```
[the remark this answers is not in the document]
Temikus: done

datastore/datastore_manager.go
```diff
@@ -17,23 +17,32 @@ package datastore
...
+	log "github.com/sirupsen/logrus"
 )
 
 // New initializes the Manager struct implementing a generic storage.Manager interface
-func New() *Manager {
-	return &Manager{}
+func New(ctx context.Context) *Manager {
+	projectID := fetchProjectID()
+
+	dsClient, err := datastore.NewClient(ctx, projectID)
```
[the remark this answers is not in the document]
Temikus: Done

datadog/metric_test.go
```diff
@@ -26,30 +26,25 @@ import (
...
+	testCtx, testCtxCancel := context.WithCancel(context.Background())
+	err := datastore.Emulator(testCtx)
 	if err != nil {
-		panic(err)
+		log.Fatalf("Unable to start the datastore emulator")
 	}
-
 	code := m.Run()
-	done()
+	testCtxCancel()
```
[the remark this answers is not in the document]
Temikus: @dparrish So as discussed I've decided to address this via returning a channel that gets pushed a message into it when it's done - PTAL

app/main.go
Temikus: Good point, no. Thank you!

app/main.go
```diff
@@ -31,20 +31,29 @@ import (
...
 	http.HandleFunc("/cleanup", cleanup)
-	appengine.Main()
+
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = "8080"
+		log.Infof("Defaulting to port %s", port)
+	}
+
+	log.Infof("Listening on http://localhost:%s", port)
```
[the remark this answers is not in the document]
Temikus: That was just a debug string for clickability, removed.

datadog/metric_test.go
```diff
@@ -188,7 +191,7 @@ func TestStackdriverDataResponses(t *testing.T) {
 			}
 			m.client.SetBaseUrl(server.URL)
 
-			desc, ts, err := m.StackdriverData(testCtx, time.Unix(1515000000, 0), &datastore.StoredMetricRecord{})
+			desc, ts, err := m.StackdriverData(testCtx, time.Unix(1515000000, 0), &datastore.StoredMetricRecord{Storage: storage})
```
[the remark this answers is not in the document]
Temikus: In this particular case - yes, it is tied to the mock query above.

app/main.go
Temikus: Ah, indeed - fixed!

datastore/emulator.go
Temikus: Unfortunately current tests require a separate env as some of them pollute the namespace which can cause issues for other tests. So it's either this or setting up some sort of cleanup routine in `testing.M` WDYT?

knyar: I think refactoring tests to allow them to run concurrently (or adding test cleanup) will be simpler than managing and restarting an emulator subprocess from Go.

datastore/datastore_manager.go
```diff
@@ -49,18 +58,32 @@ func (d *Manager) CleanupRecords(ctx context.Context, valid []string) error {
...
+	return nil
+}
+
+// ProjectID returns the name of the GCP project that code is running in.
+func fetchProjectID() string {
+	value, exists := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
+	if !exists {
+		log.Fatal("Unable to get Project ID from env, if running standalone - please set 'GOOGLE_CLOUD_PROJECT'")
+	}
+	return value
```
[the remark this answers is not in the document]
Temikus: Agreed. Changed.

datastore/emulator.go
Temikus: Chatted about this offline - agreed to keep the emulator but it's coming out or getting immediately fixed if it gets flaky.

datastore/emulator.go
Temikus: Good point. I dug quite a bit through the docs, but couldn't find specifics on child loggers, can you point at an example somewhere?

dparrish: https://godoc.org/github.com/sirupsen/logrus#New

datadog/metric.go
```diff
@@ -94,14 +91,14 @@ func (m *Metric) StackdriverData(ctx context.Context, lastPoint time.Time, rec s
 		return nil, nil, err
 	}
 	if len(series) == 0 {
-		log.Infof(ctx, "Datadog query '%s' returned no time series", m.config.Query)
+		log.WithContext(ctx).Infof("Datadog query '%s' returned no time series", m.config.Query)
```
dparrish, line 94: Just FYI, go supports a `%q` token in printf style variables, that will quote the string parameter

```go
log.WithContext(ctx).Infof("Datadog query %q returned no time series", m.config.Query)
```

It's just shorthand.

dparrish, line 94: logrus creates a child logger instance with WithContext(), so you _could_ do:

```go
l := log.WithContext(ctx)
```

At the beginning of your function, and then in the code:

```go
l.Debugf("message here")
```

`l` will always have the context attached.

datadog/metric_test.go
```diff
@@ -92,27 +86,31 @@ func makeTestServer(filename string) (*fixtureHandler, *httptest.Server) {
 }
 
 func TestStackdriverDataErrors(t *testing.T) {
+	ctx := context.Background()
+
```
dparrish, line 90: Remove this blank line

datadog/metric_test.go
```diff
@@ -26,30 +26,24 @@ import (
...
 func TestMain(m *testing.M) {
-	var done func()
-	var err error
-	testCtx, done, err = aetest.NewContext()
-	if err != nil {
-		panic(err)
-	}
-
+	ctx, cancel := context.WithCancel(context.Background())
+	// Save the emulator's quit channel
```
dparrish, line 41: Go style says that comments should be full sentences, starting with a capital and ending with a `.`

datadog/metric_test.go
```diff
@@ -92,27 +86,31 @@ func makeTestServer(filename string) (*fixtureHandler, *httptest.Server) {
...
+	storage := datastore.New(ctx)
+
 	handler, server := makeTestServer("")
 	defer server.Close()
 	m, _ := NewSourceMetric("metricname", &MetricConfig{Query: "metricquery"}, time.Second, time.Hour)
 	m.client.SetBaseUrl(server.URL)
 
 	// At this point HTTP server returns 404 to all requests, so we might as well test error handling.
-	_, _, err := m.StackdriverData(testCtx, time.Now().Add(-time.Minute), &datastore.StoredMetricRecord{})
+	_, _, err := m.StackdriverData(ctx, time.Now().Add(-time.Minute), &datastore.StoredMetricRecord{Storage: storage})
```
dparrish, line 99: You're not persisting any returned variables past this, so:

```go
if _, _, err := m.StackdriverData(...); err == nil {
  ...
}
```

datadog/metric_test.go
```diff
@@ -92,27 +86,31 @@ func makeTestServer(filename string) (*fixtureHandler, *httptest.Server) {
...
 	if err == nil {
 		t.Error("expected an error when server returns 404")
 	}
 
 	// A query needs to return a single time series.
 	handler.filename = "multiple_ts.json"
-	_, _, err = m.StackdriverData(testCtx, time.Now().Add(-time.Minute), &datastore.StoredMetricRecord{})
+	_, _, err = m.StackdriverData(ctx, time.Now().Add(-time.Minute), &datastore.StoredMetricRecord{Storage: storage})
```
dparrish, line 106: Same as line 99, do this in a single.

datadog/metric_test.go
```diff
@@ -122,6 +120,10 @@ func TestStackdriverDataErrors(t *testing.T) {
 }
 
 func TestStackdriverDataResponses(t *testing.T) {
+	ctx := context.Background()
+
```
dparrish, line 124: Remove this blank line.

datastore/datastore_manager.go
```diff
@@ -17,23 +17,30 @@ package datastore
...
+	log "github.com/sirupsen/logrus"
 )
 
 // New initializes the Manager struct implementing a generic storage.Manager interface
-func New() *Manager {
-	return &Manager{}
+func New(ctx context.Context) *Manager {
+	dsClient, err := datastore.NewClient(ctx, fetchProjectID())
+	if err != nil {
+		fmt.Errorf("could not create datastore client: %v", err)
```
dparrish, line 31: This should probably fail more catastrophically. If err is set, you don't know what dsClient will be returned.

datastore/datastore_manager.go
```diff
@@ -49,18 +56,32 @@ func (d *Manager) CleanupRecords(ctx context.Context, valid []string) error {
...
 	}
 	return nil
 }
+
+// Datastore doesn't need to be closed, this function exists for compatibility
+func (d *Manager) Close() error {
+	return nil
+}
+
+// ProjectID returns the name of the GCP project that code is running in.
```
dparrish, line 80: This comment starts with the wrong function name.

datastore/datastore_manager.go
```diff
@@ -49,18 +56,32 @@ func (d *Manager) CleanupRecords(ctx context.Context, valid []string) error {
...
+			err := d.Client.Delete(ctx, r.key(ctx))
 			if err != nil {
 				return fmt.Errorf("could not delete metric record %v: %v", r.Name, err)
 			}
 		}
 	}
 	return nil
 }
+
+// Datastore doesn't need to be closed, this function exists for compatibility
```
dparrish, line 75: Function comments must start with the name of the function.

`go vet` or `golint` should pick this up for you.

datastore/datastore_record.go
```diff
@@ -35,17 +35,20 @@ type StoredMetricRecord struct {
 
 	// CounterStartTime is used to keep start timestamp for cumulative metrics.
 	CounterStartTime time.Time
+
+	// Storage manager
```
dparrish, line 39: This comment doesn't provide any information. Remove it or fix it.

datastore/datastore_record_test.go
```diff
@@ -58,6 +46,10 @@ var metricRecordTests = []struct {
 }
 
 func TestDatastoreMetricRecords(t *testing.T) {
+	ctx := context.Background()
+
```
dparrish, line 50: Remove this blank line.

datastore/emulator.go
```diff
@@ -0,0 +1,115 @@
...
+package datastore
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	log "github.com/sirupsen/logrus"
+	"time"
+
+	//"net/http"
```
dparrish, line 24: remove

datastore/emulator.go
```diff
@@ -0,0 +1,115 @@
...
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+// Fake projectID set by the emulator
+var fakeProjectID = "testapp"
```
dparrish, line 31: If these aren't ever changed, make them `const` instead of `var`

datastore/emulator.go
```diff
@@ -0,0 +1,115 @@
...
+		fmt.Sprintf("--host-port=0.0.0.0:%v", emulatorPort),
+		"--no-store-on-disk",       // use in-memory store
+		"--consistency=1.0",        // disable random failure injection
+		"--project="+fakeProjectID, // use a fake project id so it doesn't try to infer one from environment
+	)
+	cmd.Stdout = &out
+	cmd.Stderr = &out
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true} // make this process a new group leader to help cleanup
+
+	quit := make(chan bool)
```
dparrish, line 70: You're never using the data on this channel for anything, it's purely for synchronization.
You can save a slight bit of memory by creating a channel with an empty value:

```go
quit := make(chan struct{})
```

datastore/emulator.go
```diff
@@ -0,0 +1,115 @@
...
+
+	log.Info("booting up datastore emulator...")
+	err := cmd.Start()
+	time.Sleep(2 * time.Second)
```
dparrish, line 74: Move this sleep to after the error check.
`err` will be returned immediately from the Start function. Then you can do:

```go
if err := cmd.Start(); err != nil {
  ...
}
```

datastore/emulator.go
```diff
@@ -0,0 +1,115 @@
...
+	}()
+
+	go func() {
+		select {
+		// Stop datastore emulator goroutine when context is finished
+		case <-ctx.Done():
+			cmd.Process.Signal(os.Interrupt)
+			// Give emulator time to shut down
+			time.Sleep(2 * time.Second)
+			quit <- true
```
dparrish, line 99: I think this is actively bad here. You're notifying the caller that the child has died before the child might have died. If you remove these sleep and `quit` lines, then the `cmd.Wait()` above will block until the child process has died, making this more robust.

datastore/emulator.go
```diff
@@ -0,0 +1,115 @@
...
+	ch := make(chan struct{})
+
+	go func() {
+		defer close(ch)
+		err := cmd.Wait()
+		if err != nil {
+			log.Infof("Datastore emulator killed: %v\n\t%v", err, out.String())
+		}
+		log.Info("Datastore emulator exited.")
+		quit <- true
```
dparrish, line 89: There's no need to write any data to the `quit` channel. You can just close it.

datastore/emulator.go
```diff
@@ -0,0 +1,115 @@
...
+		log.Fatalf("unable to start DS emulator: %v\n%v", err, out.String())
+		os.Exit(1)
+	}
+
+	ch := make(chan struct{})
+
+	go func() {
+		defer close(ch)
+		err := cmd.Wait()
+		if err != nil {
```
dparrish, line 85: Merge this and the previous line.

datastore/emulator.go
```diff
@@ -0,0 +1,115 @@
...
+	go func() {
+		select {
+		// Stop datastore emulator goroutine when context is finished
+		case <-ctx.Done():
+			cmd.Process.Signal(os.Interrupt)
+			// Give emulator time to shut down
+			time.Sleep(2 * time.Second)
+			quit <- true
+
+		case <-ch:
```
dparrish, line 101: You can change this to wait on the `quit` channel. That way you can remove the `ch` channel completely. You don't need 2 channels for the same notification.

influxdb/metric_test.go
```diff
@@ -271,6 +268,10 @@ func TestStackdriverDataQuery(t *testing.T) {
 }
 
 func TestStackdriverDataErrors(t *testing.T) {
+	ctx := context.Background()
+
```
dparrish, line 272: Remove this blank line.

stackdriver/adapter_test.go
```diff
@@ -17,34 +17,32 @@ package stackdriver
...
+	"github.com/google/ts-bridge/datastore"
+	"github.com/google/ts-bridge/mocks"
 	metricpb "google.golang.org/genproto/googleapis/api/metric"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
-var testCtx context.Context
+var ctx context.Context
```
dparrish, line 36: This is probably bad. You should probably never have a global variable containing a context. If you do, then I think that context is being misused

tsbridge/config.go
```diff
@@ -157,17 +154,15 @@ func NewConfig(ctx context.Context, opts *ConfigOptions) (*Config, error) {
...
 
-// projectID returns the name of the App Engine app that code is running in.
-// Empty string is returned if code is running in dev_appserver.py
-func projectID(ctx context.Context) string {
-	id := appIDFunc(ctx)
-	// dev_appserver.py returns "None" as a string. ¯\_(ツ)_/¯
-	if id == "None" {
+// projectID returns the name of the GCP project that code is running in.
+func projectID() string {
+	value, exists := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
```
dparrish, line 163: Do this in a single line:

```go
return os.Getenv("GOOGLE_CLOUD_PROJECT")
```

tsbridge/config_test.go
```diff
@@ -40,25 +44,32 @@ func TestNewConfigSimple(t *testing.T) {
...
 	}
 
 	// project_id parameter is required when app id cannot be detected.
-	for _, appid := range []string{"", "None"} {
-		appIDFunc = fakeAppIDFunc(appid)
-		_, err := NewConfig(testCtx, &ConfigOptions{Filename: "testdata/valid.yaml"})
-		if !strings.Contains(err.Error(), "please provide project_id for") {
-			t.Errorf("passing project_id should be required")
-		}
+	fakeProjectID := ""
```
dparrish, line 53: I don't understand why you're creating a local variable for an empty string, when it's never changed.

tsbridge/config_test.go
```diff
@@ -40,25 +44,32 @@ func TestNewConfigSimple(t *testing.T) {
...
+
+	setProjectID(fakeProjectID)
+	_, errNoProject := NewConfig(ctx, &ConfigOptions{Filename: "testdata/valid.yaml", Storage: storage})
+
+	if errNoProject == nil {
```
dparrish, line 58: `errNoProject` really should just be `err` for consistency.

tsbridge/metric_test.go
```diff
@@ -111,6 +111,10 @@ var metricUpdateTests = []struct {
 }
 
 func TestMetricUpdate(t *testing.T) {
+	ctx := context.Background()
+
```
dparrish, line 115: Remove this blank line.

dparrish (changes requested):

datadog/metric.go
Temikus: Didn't know about that, thanks! Changed :)

datadog/metric_test.go
Temikus: Done.

datadog/metric_test.go
Temikus: Done.

datadog/metric_test.go
Temikus: Done

datadog/metric_test.go
Temikus: Done.

datastore/datastore_manager.go
Temikus: Agreed. Done.

datastore/datastore_manager.go
Temikus: Done. Added automatic golint run to my IDE.

datastore/datastore_manager.go
Temikus: Fixed.

datadog/metric_test.go
Temikus: So genuine question here - if I do that - the function legit doesn't fit in my 27inch monitor - is it ok to leave it like this for readability sake?

datadog/metric.go
Temikus: Ah, this is awesome. I'll solve the emulator logger issue with this as well. Thanks!

datastore/datastore_record.go
Temikus: Agreed. Removed.

datastore/datastore_record_test.go
Temikus: Removed.

datastore/emulator.go
Temikus: Removed.

datastore/emulator.go
Temikus: Done.

datastore/emulator.go
Temikus: Fixed it via `log.WithContext()` PTAL

datastore/emulator.go
Temikus: Ah, this is a mistake. Fixed.

datastore/emulator.go
Temikus: Done.

datastore/emulator.go
Temikus: Removed. Do I understand correctly that my `defer close(ch)` at the beginning is enough then?

datastore/emulator.go
Temikus: Agreed. Done.

datastore/emulator.go
Temikus: Done. Removed. Agreed it's not good.

influxdb/metric_test.go
Temikus: Removed.

stackdriver/adapter_test.go
Temikus: Yeah agreed. Removed in favour of separate contexts.

datastore/emulator.go
Temikus: Changed. Thanks!

Temikus (commented):
Fixed most of it - just 2 comments outstanding - PTAL.

tsbridge/config.go
Temikus: This needs to return empty if var doesn't exist due to legacy logic in the NewConfig, I will make an effort to refactor it later though.

tsbridge/config_test.go
Temikus: Fixed.

tsbridge/config_test.go
Temikus: Question - otherwise I'll get the `No new variables on left side of :=` - is it ok to use `=` to redefine `err`?

tsbridge/metric_test.go
Temikus: Removed

tsbridge/config_test.go
dparrish: If you use the single-line form then you can redefine err! e.g. this is valid:

```go
var err error
if err := Operation(); err != nil {
...
}
```

datastore/emulator.go
```diff
@@ -0,0 +1,107 @@
...
+	cmd.Stdout = &out
+	cmd.Stderr = &out
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true} // make this process a new group leader to help cleanup
+
+	quit := make(chan struct{})
+
+	log.Info("booting up datastore emulator...")
+	if err := cmd.Start(); err != nil {
+		log.Fatalf("unable to start DS emulator: %v\n%v", err, out.String())
+		os.Exit(1)
```
dparrish, line 75: `log.Fatalf()` will crash the process, so `os.Exit(1)` will never be run

dparrish (approved):

PR #38, opened 2020-09-18
Add storage options support to datastore

Temikus:
- This PR is just a small refactor that adds `Options` struct to Datastore, abstracting the storage options initialisation, as we'll need support for additional options later, esp. for different stores that need much more options.

### DIFF
--- a/datastore/emulator.go
+++ b/datastore/emulator.go
@@ -19,6 +19,9 @@ import (
 	"context"
 	"fmt"
 	log "github.com/sirupsen/logrus"
+	"io/ioutil"
+	"net"
+	"strings"
 	"time"
 
 	"os"
@@ -28,9 +31,6 @@ import (
 
 // Fake projectID set by the emulator
 const fakeProjectID = "testapp"
-
-// Port for emulator to use
-var emulatorPort = 30801
 
 // Emulator instantiates datastore emulator for use in tests.
 //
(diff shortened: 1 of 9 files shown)


dparrish (approved):

datastore/emulator.go
```diff
@@ -104,3 +116,31 @@ func Emulator(ctx context.Context) <-chan struct{} {
...
 	return quit
 }
+
+func findEmulatorPath() string {
+	gcloudPath, err := exec.LookPath("gcloud")
+	if err != nil {
+		log.Fatalf("Couldn't determine gcloud path")
+	}
+
+	emulatorPath := strings.Replace(gcloudPath, "bin/gcloud", "platform/cloud-datastore-emulator/cloud_datastore_emulator", 1)
```
dparrish, line 126: Go style encourages inlining instead of creating single-use variables.

```
return strings.Replace(gcloudPath, "bin/gcloud", "platform/cloud-datastore-emulator/cloud_datastore_emulator", 1)
```

datastore/emulator.go
```diff
@@ -104,3 +116,31 @@ func Emulator(ctx context.Context) <-chan struct{} {
 
 	return quit
 }
+
+func findEmulatorPath() string {
+	gcloudPath, err := exec.LookPath("gcloud")
```
dparrish, line 121: Go style encourages short variable names for small scopes. `path` or even `p` would be preferable here.

datastore/emulator.go
```diff
@@ -53,23 +53,35 @@ func Emulator(ctx context.Context) <-chan struct{} {
 	// Set log formatter to process newlines since emulator gives a lot of output
 	l.Logger.SetFormatter(&log.TextFormatter{DisableQuote: true})
 
-	// Set up the command
+	// Set up
```
dparrish, line 56: Go style says that all comments should be full sentences, starting with a capital and ending with a full stop.
If that's too much comment for a single line of code, then there shouldn't be a comment for it.

dparrish (changes requested):

datastore/emulator.go
Temikus: Done

datastore/emulator.go
Temikus: Agreed, this is a bad leftover. Fixed.

datastore/emulator.go
Temikus: Done.

PR #40, opened 2020-09-30
Adding CLI flags & 12-factor config

Temikus:
- getting rid of all env getters and validations and putting them into one single place
- doing simple flag-based logic to set Env instead of importing a large dep like Kong/Kingpin/Viper
	- there is 2 cons to the approach but I consider them acceptable:
		- some extra/code plumbing needed (see `flags.go`)
		- env will override flags if set

BREAKING 💥: if we don't want option to be set we don't set it at all instead of setting it to empty. I think it's worth the possible light breakage as this makes all env options explicit, instead of using brittle check-if-empty logic.

@Dnefedkin FYI

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -63,7 +63,7 @@ func sync(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithTimeout(ctx, t)
 	defer cancel()
 
-	if isAppEngine() && r.Header.Get("X-Appengine-Cron") != "true" {
+	if env.IsAppEngine() && r.Header.Get("X-Appengine-Cron") != "true" {
 		http.Error(w, "Only cron requests are allowed here", http.StatusUnauthorized)
 		return
 	}
@@ -116,7 +116,7 @@ func sync(w http.ResponseWriter, r *http.Request) {
 func cleanup(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
-	if isAppEngine() && r.Header.Get("X-Appengine-Cron") != "true" {
+	if env.IsAppEngine() && r.Header.Get("X-Appengine-Cron") != "true" {
 		http.Error(w, "Only cron requests are allowed here", http.StatusUnauthorized)
 		return
 	}
@@ -216,7 +216,7 @@ func loadStorageEngine(ctx context.Context) (storage.Manager, error) {
 		datastoreManager := datastore.New(ctx, &datastore.Options{})
 		return datastoreManager, nil
 	case "boltdb":
-		if isAppEngine() {
+		if env.IsAppEngine() {
 			return nil, fmt.Errorf("BoltDB storage is not supported on AppEngine")
 		}
 		opts := &boltdb.Options{DBPath: os.Getenv("BOLTDB_PATH")}
@@ -229,10 +229,4 @@ func loadStorageEngine(ctx context.Context) (storage.Manager, error) {
 	default:
 		return nil, fmt.Errorf("unknown storage engine selected: %s", storageEngine)
 	}
-}
-
-// Check if we're running in AppEngine by checking GAE_ENV variable
-func isAppEngine() bool {
-	_, set := os.LookupEnv("GAE_ENV")
-	return set
 }
--- a/app/main.go
+++ b/app/main.go
@@ -19,48 +19,110 @@ import (
 	"errors"
 	"fmt"
 	"html/template"
+	"net"
 	"net/http"
-	"os"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/google/ts-bridge/boltdb"
 	"github.com/google/ts-bridge/datastore"
+	"github.com/google/ts-bridge/env"
 	"github.com/google/ts-bridge/stackdriver"
 	"github.com/google/ts-bridge/storage"
 	"github.com/google/ts-bridge/tsbridge"
 
 	"github.com/dustin/go-humanize"
 	log "github.com/sirupsen/logrus"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+	debug = kingpin.Flag("debug", "enable debug mode").Envar("DEBUG").Default("false").Bool()
+	port  = kingpin.Flag("port", "ts-bridge server port").Envar("PORT").Default("8080").Int()
+
+	metricConfig = kingpin.Flag(
+		"metric-config", "metric configuration file path",
+	).Envar("CONFIG_FILE").Default("metrics.yaml").String()
+
+	enableStatusPage = kingpin.Flag(
+		"enable-status-page", "enable ts-bridge server status page",
+	).Envar("ENABLE_STATUS_PAGE").Default("false").Bool()
+
+	updateTimeout = kingpin.Flag(
+		"update-timeout", "total timeout for updating all metrics.",
+	).Envar("UPDATE_TIMEOUT").Default("5m").Duration()
+
+	updateParallelism = kingpin.Flag(
+		"update-parallelism", "number of metrics to update in parallel",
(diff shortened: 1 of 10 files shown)


app/flags.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"os"
+	"strconv"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func reconcileEnvBool(key string, defaultVal bool) bool {
+	if _, set := os.LookupEnv(key); set {
```
dparrish, line 14: I worry that this logic is subtly confusing. I would expect that a command-line flag would override an environment variable, not the other way around.

I envision the "WTF IS GOING ON?!" moment as I run the code and my command-line flag changes don't take effect at all because something is hidden in the environment.

dparrish (changes requested):

app/main.go
```diff
@@ -19,34 +19,83 @@ import (
...
 
 	"github.com/dustin/go-humanize"
 	log "github.com/sirupsen/logrus"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+	debug = kingpin.Flag(
+		"debug", "enable debug mode",
+	).Envar("DEBUG").Default("false").Bool()
```
dparrish, line 42: Go style doesn't have a line length restriction, which avoids having to do this kind of line breaking making the code longer and (IMHO) less readable. Go style prefers to minimize height.

app/main.go
```diff
@@ -81,31 +125,25 @@ func sync(w http.ResponseWriter, r *http.Request) {
...
-	p, err := strconv.Atoi(os.Getenv("UPDATE_PARALLELISM"))
-	if err != nil {
-		logAndReturnError(ctx, w, fmt.Errorf("could not parse UPDATE_PARALLELISM: %v", err))
-		return
-	}
-	if p < 1 || p > 100 {
-		logAndReturnError(ctx, w, fmt.Errorf("expected UPDATE_PARALLELISM between 1 and 100; got %d", p))
-		return
+	if *updateParallelism < 1 || *updateParallelism > 100 {
+		logAndReturnError(ctx, w, fmt.Errorf("expected UPDATE_PARALLELISM between 1 and 100; got %d", *updateParallelism))
```
dparrish, line 143: Can you do this check in `main` so that the process won't start if the value is wrong?

As it is, it'll return OK to the health check but never be able to handle a query if this argument is wrong.

datastore/datastore_manager.go
```diff
@@ -26,11 +26,23 @@ import (
...
-	dsClient, err := datastore.NewClient(ctx, fetchProjectID())
+	if options.Project == "" {
+		if env.IsAppEngine() {
+			options.Project = env.AppEngineProject()
+			log.Infof("No datastore project specified, defaulting to GAE project: %v", options.Project)
+		} else {
+			log.Fatalf(
+				"Could not determine project to use for Datastore, please set DATASTORE_PROJECT or --datastore-project flag")
+		}
+	}
```
dparrish, line 43: Since Go style recommends minimizing vertical space and indent levels, switch the logic around so that there's no `else` block:

```go
if options.Project == "" {
  if !env.IsAppEngine() {
    log.Fatalf("Could not determine project to use for Datastore, please set DATASTORE_PROJECT or --datastore-project flag")
  }
  options.Project = env.AppEngineProject()
  log.Infof("No datastore project specified, defaulting to GAE project: %v", options.Project)
}
```

Also, no need to line wrap the log.Fatalf line.

datastore/emulator.go
```diff
@@ -108,7 +108,12 @@ func Emulator(ctx context.Context) <-chan struct{} {
 	}
 
 	// This is emulating the old `aetest` framework that has a hardcoded "testapp" value for project
-	if err := os.Setenv("GOOGLE_CLOUD_PROJECT", fmt.Sprintf(fakeProjectID)); err != nil {
+	if err := os.Setenv("GOOGLE_CLOUD_PROJECT", fakeProjectID); err != nil {
+		log.Errorf("couldn't set env GOOGLE_CLOUD_PROJECT: %v", err)
```
dparrish, line 112: What will happen to the code that follows this? Will everything fail? Should this be a `log.Fatalf` instead?

datastore/emulator.go
```diff
@@ -108,7 +108,12 @@ func Emulator(ctx context.Context) <-chan struct{} {
...
+	}
+
+	// This env is set to pretend we're running on AppEngine
+	if err := os.Setenv("GAE_ENV", "standard"); err != nil {
```
dparrish, line 116: Check is for GAE_ENV, error is for GOOGLE_CLOUD_PROJECT

tsbridge/stats.go
```diff
@@ -56,16 +58,18 @@ type StatsCollector struct {
...
-	}
 	if project == "" {
-		return nil, fmt.Errorf("Please define SD_PROJECT_FOR_INTERNAL_METRICS if not running on App Engine")
+		if env.IsAppEngine() {
+			project = os.Getenv("GOOGLE_CLOUD_PROJECT")
+			log.Infof("Cannot determine project to store stats in, defaulting to GAE project: %v", project)
+		} else {
+			return nil, fmt.Errorf("error initializing stats collector - project empty: set SD_PROJECT_FOR_INTERNAL_METRICS or --stats-sd-project if not running on App Engine")
+		}
+
```
dparrish, line 72: As above, flip the logic order so there's no need for the `else` block.

dparrish (changes requested):
I think this is looking much cleaner now. There's definitely less code for the same outcome, and the encapsulation is much tighter. Just a few cleanups to go.

app/main.go
Temikus: Added a flag validator.

Temikus: Added a flag validator.

datastore/datastore_manager.go
Temikus: Agreed on line wrap.

I would prefer leaving the env.IsAppengine however to follow the "positive boolean" pattern: https://engdoc.corp.google.com/eng/doc/tott/episodes/530.md?cl=head

WDYT?

datastore/emulator.go
Temikus: Agreed, changed to Fatal.

datastore/emulator.go
Temikus: Shame on me T__T Fixed.

tsbridge/stats.go
Temikus: Ditto on https://engdoc.corp.google.com/eng/doc/tott/episodes/530.md?cl=head

app/main.go
Temikus: I do agree it blurs a bit, how about I space those out?

I'm hesitant to make them one liners as `counterResetInterval` for example is 217 characters long and doesn't fit on my 31" screen, which makes it not very readable. Or maybe I'm missing something here?

Temikus (commented):
Addressed all comments, but have some questions. PTAL.

app/main.go
dparrish: That might be a reasonable one to wrap individually. That doesn't mean all of these need to be wrapped.

datastore/datastore_manager.go
dparrish: I see the reason. Just know if this ever goes to Go Readability approval team, they'll insist on it being the other way around.

dparrish (approved):

datastore/datastore_manager.go
Temikus: Ack. This is good to know. Thanks!

app/main.go
Temikus: Gotcha. Unwrapped everything that's less than 120 lines and added spaces between large options for readability.

PR #41, opened 2020-10-02
Set up a sync loop for standalone use, refactor to unload main

Temikus:
This PR
- Refactors the configuration logic by separating config into two proper domains - MetricConfig and Config, disambiguating the two
- Refactors http endpoints into actions and wrappers to be able to run them standalone
- Adds a sync/cleanup goroutine and configuration for it

@knyar interested in your thoughts if you have a bit of time to spare (there will be a separate readability review, more interested in your thoughts on the approach)

@Dnefedkin FYI


app/main.go
```diff
@@ -72,6 +73,14 @@ var (
 		"stats-sd-project", "Stackdriver project for internal ts-bridge metrics",
 	).Envar("SD_PROJECT_FOR_INTERNAL_METRICS").String()
 
+	syncPeriod = kingpin.Flag(
+		"sync-period", "How often to sync metrics when running in standalone mode",
+	).Envar("SYNC_PERIOD").Default("60s").Duration()
+
+	syncCleanupAfter = kingpin.Flag(
+		"sync-cleanup-after", "Run cleanup after X sync loops",
+	).Envar("SYNC_CLEANUP_AFTER").Default("100").Int()
```
knyar, line 82: I'd suggest making this a Duration as well.

tsbridge/metric_config.go
```diff
@@ -0,0 +1,157 @@
...
+			return nil, err
+		}
+	}
+
+	log.WithContext(ctx).Debugf("Read %d metrics and %d destinations from the config file", len(metrics), len(destinations))
+	return c, nil
+}
+
+// projectID returns the name of the GCP project that code is running in.
+func projectID() string {
```
knyar, line 151: Do you need this now when you have `env.AppEngineProject()`?

app/main.go
```diff
@@ -96,9 +105,62 @@ func main() {
...
+	})
+	http.HandleFunc("/sync", func(w http.ResponseWriter, r *http.Request) {
+		syncHandler(w, r, config)
+	})
+	http.HandleFunc("/cleanup", func(w http.ResponseWriter, r *http.Request) {
+		cleanupHandler(w, r, config)
+	})
+
+	// Run a sync loop for standalone use
+	if env.NotAppEngine() {
```
knyar, line 132: I think a separate function that does negation just makes the logic more confusing. `if !env.IsAppEngine()` is just as readable and short.

app/main.go
```diff
@@ -96,9 +105,62 @@ func main() {
...
+		go func() {
+			defer cancel()
+			for {
+				select {
+				case <-time.After(config.Options.SyncPeriod):
+					ctx, cancel := context.WithTimeout(ctx, *updateTimeout)
+					log.WithContext(ctx).Debugf("Running sync...")
+					if err := sync(ctx, config); err != nil {
+						log.WithContext(ctx).Errorf("error running sync() routine: %v", err)
+						return
```
knyar, line 145: What's the strategy behind returning here? It seems like this will stop the sync loop, but the process will keep running.

app/main.go
```diff
@@ -96,9 +105,62 @@ func main() {
...
+	// Run a sync loop for standalone use
+	if env.NotAppEngine() {
+		log.Debug("Running outside of appengine, starting up a sync loop...")
+		ctx, cancel := context.WithCancel(context.Background())
+		count := 0
+		go func() {
+			defer cancel()
+			for {
+				select {
+				case <-time.After(config.Options.SyncPeriod):
```
knyar, line 140: Won't this make actual interval between sync operations longer than SyncPeriod (given that running sync takes non-zero amount of time)?

app/main.go
```diff
@@ -96,9 +105,62 @@ func main() {
...
+					ctx, cancel := context.WithTimeout(ctx, *updateTimeout)
+					log.WithContext(ctx).Debugf("Running sync...")
+					if err := sync(ctx, config); err != nil {
+						log.WithContext(ctx).Errorf("error running sync() routine: %v", err)
+						return
+					}
+					// perform a cleanup every 100th cycle
+					if count%config.Options.SyncCleanupAfter == 0 {
```
knyar, line 148: Perhaps using a separate timer for cleanup will be simpler.

app/main.go
```diff
@@ -96,9 +105,62 @@ func main() {
...
+		SDLookBackInterval:       *sdLookBackInterval,
+		SDInternalMetricsProject: *sdInternalMetricsProject,
+		UpdateParallelism:        *updateParallelism,
+		EnableStatusPage:         *enableStatusPage,
+		StorageEngine:            *storageEngine,
+		SyncPeriod:               *syncPeriod,
+		SyncCleanupAfter:         *syncCleanupAfter,
+	})
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
```
knyar, line 121: Instead of wrapping each handler in a function, perhaps you could attach `config` to a struct and make each handler a struct method?

tsbridge/config.go
```diff
@@ -1,168 +1,29 @@
...
-// InfluxDBMetricConfig combines common metric configuration parameters with InfluxDB-specific ones.
-type InfluxDBMetricConfig struct {
-	SourceMetricConfig    `yaml:"_,inline"`
-	influxdb.MetricConfig `yaml:"_,inline"`
-}
-
-// Metrics returns a list of metrics defined in the configuration file.
-func (c *Config) Metrics() []*Metric {
-	return c.metrics
+	Options ConfigOptions
```
knyar, line 8: Does Config need to be a separate struct from ConfigOptions? It seems unnecessary given that one only contains the other.

knyar (changes requested):

app/main.go
Temikus: There are sometimes transient errors with running this routine (issues on Datadog or SD side for example).

This is a good point though - should I introduce a backoff counter here? E.g. exit the loop if it failed more than 3 times.

app/main.go
Temikus: Duration here is not really appropriate as the amount of things to clean up is not proportional to time, rather to amount of sync loops, so I would leave it as is. Do you think that's acceptable?

tsbridge/metric_config.go
Temikus: Agreed, fixed. Just had to change the `env` module a little bit to make sure we don't run into confusion issues in the future should AppEngine change env.

app/main.go
Temikus: This was added for another purpose which is no longer relevant, so agreed. Removing.

app/main.go
Temikus: Indeed, but I think it's an acceptable tradeoff between this and and spawning an extra routine every 10 minutes as that can cause race conditions on large amount of metrics (or issues + backoff that take >10m) unless we complicate this with locks.

WDYT?

app/main.go
Temikus: Not really since BoltDB supports only 1 writer at the time. Then we need to use locks to prevent race conditions and do a lot of brittle timing, no?

tsbridge/config.go
Temikus: This is intentional as I later want to refactor how storage configuration works, so I'm keeping all flags in the same place.

app/main.go
Temikus: Good point, refactored. I've also moved all web ui/tasks stuff into a separate package. PTAL.

Temikus:
@knyar Thank you for the review! Addressed your comments. PTAL when you have a chance.

app/main.go
knyar: I think if users ask us to sync every 2 minutes, we should sync every 2 minutes; not 2+X minutes where X is outside of user's control).

There definitely should never be more than 1 syncing process running (since it could create out of order writes to Stackdriver), so I think the options here are:
1. measure how long syncing took, and then dynamically calculate how long to sleep for until the next sync (i.e. `sleep(sync_period - latency)`)
2. kick off syncing regularly, but check whether there is an existing sync running (a lock might be the easiest way to implement this). If there is an ongoing sync, just do nothing until the next timer. This is similar to how App Engine cron works, I believe.

app/main.go
knyar: I am not sure I understand what would exiting the loop achieve if the process will keep running.

I don't think you need backoff here given how rarely syncing happens (minutes rather than seconds). I think logging an error and trying again during the next scheduled sync is sufficient.

env/env.go
```diff
@@ -10,8 +10,14 @@ func IsAppEngine() bool {
...
+// 	 Note: this is separate from AppEngine project since other tools/services use this env var
+func GoogleCloudProject() string {
+	return os.Getenv("GOOGLE_CLOUD_PROJECT")
+}
+
 // AppEngineProject returns the cloud project GAE App is running in
 // 	 Note: this is the official way to get this information within GAE.
 func AppEngineProject() string {
-	return os.Getenv("GOOGLE_CLOUD_PROJECT")
+	return GoogleCloudProject()
```
knyar, line 22: I don't really understand why you need a one-line function that calls another functions. Perhaps just call `GoogleCloudProject()` directly?

web/tasks.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"context"
+	"errors"
+	"strings"
+
+	"github.com/google/ts-bridge/stackdriver"
+	"github.com/google/ts-bridge/tsbridge"
+)
+
+// Sync updates all configured metrics.
```
knyar, line 12: I am not sure Sync and Cleanup belong to the `web` package.

app/main.go
```diff
@@ -96,9 +98,61 @@ func main() {
...
+		EnableStatusPage:         *enableStatusPage,
+		StorageEngine:            *storageEngine,
+		SyncPeriod:               *syncPeriod,
+		SyncCleanupAfter:         *syncCleanupAfter,
+	})
+
+	h := web.NewHandler(config)
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		web.Index(w, r, config)
```
knyar, line 118: Should `Index` just be part of `web.Handler` as well? I don't think it gets called independently now.

knyar (changes requested):
Thanks for moving the web handlers into a separate package, it looks much cleaner now.

app/main.go
Temikus: Yeah, I think I prematurely optimised (in case we needed to do some other DB maintenance).

Set up the cleanup routine to run on startup. PTAL.

app/main.go
Temikus: Oh ok, no, this is just my mistake as I didn't intend to quit the main loop. Fixed.

env/env.go
Temikus: This is an alias, essentially. Just `GOOGLE_CLOUD_PROJECT` set doesn't necessarily mean AppEngineEnvironment because we use the same env vars for tooling _and_ runtimes.

So I want the code to be explicit that "here we need to get the AppEngine project" vs "here we need to see if the env var is set"

Make sense?

app/main.go
Temikus: Fixed

web/tasks.go
Temikus: Moved into a `tasks` package

app/main.go
Temikus: Gotcha. I've played around a bit with a couple of variants and couldn't find a simple one. Since async logic and timing are very constly in terms of maintenance and debugging, I've decided to add a TODO in here which I plan to resolve in another PR with a whole bunch of testing around it.

app/main.go
```diff
@@ -96,9 +95,53 @@ func main() {
...
+	h := web.NewHandler(config)
+
+	http.HandleFunc("/", h.Index)
+	http.HandleFunc("/sync", h.Sync)
+	http.HandleFunc("/cleanup", h.Cleanup)
+
+	// Run a cleanup on startup
+	log.Debugf("Performing startup cleanup...")
+	if err := tasks.Cleanup(context.Background(), config); err != nil {
+		log.Errorf("error running the Cleanup() routine: %v", err)
```
knyar, line 120: How do you intend users to detect and respond to this? Perhaps `log.Exitf` would be more appropriate?

env/env.go
knyar: Sorry, I don't think I understand the difference between 'Google Cloud project' and 'App Engine project'.

This environment variable gets set by App Engine automatically and corresponds to the GCP project that the app belongs to. It is used as the default value for the following configuration options:
- Datastore project for metric records (unless `DATASTORE_PROJECT` is set)
- Stackdriver project for internal stats (unless `SD_PROJECT_FOR_INTERNAL_METRICS` is set)
- Stackdriver project for collected metrics (unless stackdriver destination has a `project_id` field set)

I see the value in adding a separate function to read the environment variable instead of calling `os.Getenv` several times, but I don't understand why we need two separate functions for this.

web/common.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"context"
+	"net/http"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// Since some URLs are triggered by App Engine cron, error messages returned in HTTP response
+// might not be visible to humans. We need to log them as well, and this helper function does that.
+func logAndReturnError(ctx context.Context, w http.ResponseWriter, err error) {
```
knyar, line 12: nit: Do you really need a separate file for this function?

knyar (approved):
Just a few minor comments, but this lgtm overall. I'd suggest asking someone from CRE to review as well.

web/common.go
Temikus: Now that I pulled the other function out, no. Moved to handlers.

app/main.go
Temikus: Agreed, changed to `log.Fatalf()`

env/env.go
Temikus: So I wanted to make sure we don't mix up intentional use of it inside some tool outside of AppEngine. But now that I think about it - we haven't really had this thing run outside AppEngine so it doesn't make sense to separate this. Sorry for all the confusion on this one :|

env/env.go
```diff
@@ -10,8 +10,14 @@ func IsAppEngine() bool {
 	return set
 }
 
+// Extract google cloud project from environment
```
nerdinary: Go comments must start with the func name.

i.e, // GoogleCloudProject extracts ths ....

tasks/tasks.go
```diff
@@ -0,0 +1,88 @@
+package tasks
```
nerdinary, line 1: Could you add a top level docstring to describe this package?

web/handlers.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		logAndReturnError(ctx, w, err)
+		return
+	}
+	if err := t.Execute(w, metrics.Metrics()); err != nil {
+		logAndReturnError(ctx, w, err)
+	}
+}
+
+// Since some URLs are triggered by App Engine cron, error messages returned in HTTP response
```
nerdinary, line 89: `// logAndReturnError ....`

app/main.go
```diff
@@ -117,6 +128,39 @@ func main() {
 		cleanupHandler(w, r, config)
 	})
 
+	// Run a sync loop for standalone use
+	if !env.IsAppEngine() {
+		log.Debug("Running outside of appengine, starting up a sync loop...")
+		ctx, cancel := context.WithCancel(context.Background())
+		count := 0
+		go func() {
```
nerdinary, line 128: readability nit:

Can you move this to it's own func to make it easier to read?

env/env.go
Temikus: Already fixed in the recent rebase 👍

tasks/tasks.go
Temikus: Added.

tasks/tasks.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/google/ts-bridge/boltdb"
+	"github.com/google/ts-bridge/datastore"
+	"github.com/google/ts-bridge/env"
+	"github.com/google/ts-bridge/stackdriver"
+	"github.com/google/ts-bridge/storage"
+	"github.com/google/ts-bridge/tsbridge"
+	"strings"
+)
+
+// Helper function to load the correct storage manager depending on settings
```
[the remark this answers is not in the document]
Temikus: Done.

tasks/tasks.go
```diff
@@ -0,0 +1,88 @@
...
+func LoadStorageEngine(ctx context.Context, config *tsbridge.Config) (storage.Manager, error) {
+	switch config.Options.StorageEngine {
+	case "datastore":
+		datastoreManager := datastore.New(ctx, &datastore.Options{Project: config.Options.DatastoreProject})
+		return datastoreManager, nil
```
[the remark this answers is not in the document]
Temikus: Yes it is meant to be expanded in the future as this is where we can pipe in any client options, so I'll leave this particular case as-is.

tasks/tasks.go
```diff
@@ -0,0 +1,88 @@
...
+	case "boltdb":
+		if env.IsAppEngine() {
+			return nil, fmt.Errorf("BoltDB storage is not supported on AppEngine")
```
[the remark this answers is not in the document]
Temikus: Ta! Fixed.

web/handlers.go
```diff
@@ -0,0 +1,94 @@
+package web
```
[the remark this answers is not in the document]
Temikus: Done!

web/handlers.go
Temikus: Fixed!

app/main.go
```diff
@@ -109,6 +137,25 @@ func main() {
 
 }
 
+func syncLoop(cancel context.CancelFunc, config *tsbridge.Config, ctx context.Context) {
+	func() {
```
nerdinary, line 141: You shouldn't need a `func()` here, since you're already calling `syncLoop` in a goroutine.

app/main.go
```diff
@@ -96,9 +95,38 @@ func main() {
...
+	if err := tasks.Cleanup(context.Background(), config); err != nil {
+		log.Fatalf("error running the Cleanup() routine: %v", err)
+	}
+
+	// Run a sync loop for standalone use
+	// TODO(temikus): refactor this to run exactly every SyncPeriod and skip sync if one is already active
+	if !env.IsAppEngine() {
+		log.Debug("Running outside of appengine, starting up a sync loop...")
+		ctx, cancel := context.WithCancel(context.Background())
+		go syncLoop(cancel, config, ctx)
```
nerdinary, line 128: readability: convention is to always pass context first.

`go syncLoop(ctx, cancel, config)`

app/main.go
Temikus: D'oh, fixed!

app/main.go
Temikus: 👍 Good to know, fixed!

app/main.go
Temikus: Done.

PR #43, opened 2020-10-20
Add a simple health check endpoint

Temikus:
This adds a simple health check endpoint to TS-Bridge

Currently it just returns `status: ok` (i.e. a simple heartbeat that the binary is running) but can be extended later. This seems to be enough for now as we fatally exit on all conditions that I know of that could cause an application to stall, but the binary to keep running.

FYI @Dnefedkin


nerdinary (approved):
LGTM.

PR #48, opened 2020-11-03
Roll static assets into the binary

Temikus:
This makes deployments easier and makes sure that the assets are always versioned with the binary.

Part of containerisation work.


nerdinary (approved):

Temikus:
It looks like `packr2` uses absolute paths to generate the files so it will not produce reproducible builds :| We can ignore those files and build them from scratch every time but I'm not a fan of a setup like this tbh since it means 2 builds in 2 different directories will have different checksums.

I'm abandoning this PR in favour of a new one which I'll work on later this week.

PR #49, opened 2020-11-05
Static assets

Temikus:
Second attempt at static assets bundling, this time with no bundling extra deps with the package.

TODO added to replace with embed.FS when it is released as a part of Go stdlib


nerdinary (commented):
"go-bindata" to generate the static data is causing travis ci to fail.

nerdinary (approved):

Issue #42, opened 2020-10-12
Fix TSBridge UserAgent presentation


Temikus:
Currently TSBridge just sends requests with standard `google-cloud-go` user agent, leaving the application bit empty.

Let's pick user agent from packer:
google-api-go-client/0.5 Packer/1.6.3 (+https://www.packer.io/; go1.15.2; linux/amd64),gzip(gfe)

```
google-api-go-client/%VERSION% ts-bridge/%VERSION% (go%VERSION%; %PLATFORM%)
```
, e.g.:
```
google-api-go-client/0.5 ts-bridge/0.1 (go1.13.2; linux/amd64)
```

Temikus:
@nerdinary Adi, PTAL

PR #50, opened 2020-11-05
Add a version package that records version revisions, along with helpers for adding a useragent.

nerdinary:


nerdinary:
Fixes #42

version/version.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"google.golang.org/api/googleapi"
+)
+
+const (
+	// Project is the name of this project.
+	Project = "ts-bridge"
+
+	// Ver is the current version of this project.
+	Ver = "0.1"
```
Temikus, line 17: We need to adhere to [semantic versioning](https://semver.org/), so this needs to be "0.1.0"

stackdriver/metric_client.go
```diff
@@ -18,7 +18,9 @@ import (
 	"context"
 
 	monitoring "cloud.google.com/go/monitoring/apiv3"
+	"github.com/google/ts-bridge/version"
```
Temikus, line 21: nit: please separate internal imports and external imports, see: https://github.com/google/ts-bridge/blob/c0e8016838f2da717e4c672cf1bd8220fb9ff73f/app/main.go#L28

version/version_test.go
```diff
@@ -0,0 +1,16 @@
+package version
+
+import "testing"
+
+func TestUserAGent(t *testing.T) {
+	rt = "go1.2.3"
+	goos = "linux"
+	goarch = "amd64"
+	version = "0.1"
+	googClient = "googleapi-go/0.1"
```
Temikus, line 10: Nit: this is not the line google api agent actually has, usually. 

Since google api libs are prone to changes, is it possible to do use the live import here + a regex match?

E.g. "google-api-go-client/[\.0-9]" or something?

version/version.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// goarch is the architechture.
+	goarch = runtime.GOARCH
+
+	// googClient is the google API client version.
+	googClient = googleapi.UserAgent
+)
+
+// Revision returns the current revision of this project.
+func Revision() string {
```
Temikus, line 39: I know this is slightly out of scope, but can I ask you to pipe it to `main.go` so it prints the revision if `--version` flag is passed?

We will need it for CI/releasing.

Feel free to push back if you want to make this a separate task.

version/version_test.go
Temikus: That way we'll be alerted if the agent changes.

Temikus (changes requested):
This is great, just some minor comments. PTAL.

version/version.go
nerdinary: Done.

version/version_test.go
nerdinary: Yup, makes sense. Updated.

version/version.go
Temikus: Not quite. Sorry, my bad. Basically we need the standard UNIX behaviour here. If `--version` flag is passed - print the version and exit immediately with 0.

E.g.:
```
λ fly --version
5.6.0
λ echo $?
0
```
This is needed for packaging automation so the version in the app matches everything else.

Temikus (approved):
LGTM, just fix the `version` behaviour.

Temikus:
Note: please squash when merging.

version/version.go
nerdinary: Done.

Issue #52, opened 2020-11-13
Investigate memory usage


Temikus:
Looks like we may have a memory leak, some investigation is required with a profiler: ![6jCwjZXQGp4hgNQ](https://user-images.githubusercontent.com/2083229/99034267-e75d2a80-25d0-11eb-9ca1-3125188cf5ca.png)

Temikus:
Looks ***much*** better now:
<img width="1112" alt="Screen Shot 2020-11-19 at 10 12 39 am" src="https://user-images.githubusercontent.com/2083229/99599637-d8063300-2a4f-11eb-81d2-333dc5cc35b6.png">

PR #56, opened 2020-11-18
Fixing the goroutine leak

Temikus:
Plus improved memory usage by instantiating configs, etc. only once.

Additionally:
- Bump to 0.1.1
- Add Google Cloud Profiler support

Fixing #52

### DIFF
--- a/tasks/tasks.go
+++ b/tasks/tasks.go
@@ -5,13 +5,23 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
+	"sync"
+
 	"github.com/google/ts-bridge/boltdb"
 	"github.com/google/ts-bridge/datastore"
 	"github.com/google/ts-bridge/env"
 	"github.com/google/ts-bridge/stackdriver"
 	"github.com/google/ts-bridge/storage"
 	"github.com/google/ts-bridge/tsbridge"
-	"strings"
+)
+
+var (
+	sdClient     *stackdriver.Adapter
+	sdClientOnce sync.Once
+
+	statsCollector     *tsbridge.StatsCollector
+	statsCollectorOnce sync.Once
 )
 
 // LoadStorageEngine is a helper function to load the correct storage manager depending on settings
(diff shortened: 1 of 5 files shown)


tasks/tasks.go
```diff
@@ -40,24 +57,28 @@ func Sync(ctx context.Context, config *tsbridge.Config) error {
...
 	defer store.Close()
 
-	metrics, err := tsbridge.NewMetricConfig(ctx, config, store)
-	if err != nil {
-		return err
-	}
+	metricConfigOnce.Do(func() {
+		metricConfig, err = tsbridge.NewMetricConfig(ctx, config, store)
+		if err != nil {
+			log.Fatalf("Unable to initialize metric config: %v", err)
```
nerdinary, line 63: hmm, I'm not sure log.Fatal is the right call here, as it'll call os.Exit right after, and does not allow for clean closure of open handlers.

Might be cleaner to return an error instead, which allows for handlers such as `store.Close()` etc to be executed.

Same for the other log.Fatal's below.

tasks/tasks.go
```diff
@@ -11,7 +11,24 @@ import (
 	"github.com/google/ts-bridge/stackdriver"
 	"github.com/google/ts-bridge/storage"
 	"github.com/google/ts-bridge/tsbridge"
+	"log"
 	"strings"
+	"sync"
+)
+
+var (
```
nerdinary, line 19: readability:

convention is to have a single var declaration, instead of multiple groups.

```
var (
	sdClient     *stackdriver.Adapter
	sdClientOnce sync.Once

	metricConfig     *tsbridge.MetricConfig
	metricConfigOnce sync.Once

	statsCollector     *tsbridge.StatsCollector
	statsCollectorOnce sync.Once
)
```

nerdinary (approved):

tasks/tasks.go
Temikus: That's a good point. Fixed. Can you PTAL?

tasks/tasks.go
```diff
@@ -5,13 +5,25 @@ import (
...
+	sdClient     *stackdriver.Adapter
+	sdClientOnce sync.Once
+
+	statsCollector     *tsbridge.StatsCollector
+	statsCollectorOnce sync.Once
+
+	errSync []error
```
nerdinary, line 26: nit: errors should be localized to the func that's returning it.

I'd suggest creating this var right before you call `sdClientOnce.Do(func()` instead of a global var.

tasks/tasks.go
```diff
@@ -45,19 +57,26 @@ func Sync(ctx context.Context, config *tsbridge.Config) error {
...
 
-	sd, err := stackdriver.NewAdapter(ctx, config.Options.SDLookBackInterval)
-	if err != nil {
-		return err
-	}
-	defer sd.Close()
+	sdClientOnce.Do(func() {
+		sdClient, err = stackdriver.NewAdapter(ctx, config.Options.SDLookBackInterval)
+		if err != nil {
+			errSync = append(errSync, fmt.Errorf("Unable to initialize stackdriver adapter: %v", err))
```
nerdinary, line 63: readability nit: error strings should not be capitalized. same for other error strings in this func.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

nerdinary (approved):

tasks/tasks.go
Temikus: Agreed. Moved.

tasks/tasks.go
Temikus: Fixed!

PR #64, opened 2020-12-15
Cache cloudbuild and specify machine type in Cloud Build

SoapHia:
- Add cache to cloudbuild and specify machine type to speed up the build process. ([Documentation for reference])(https://cloud.google.com/cloud-build/docs/speeding-up-builds)
- Change the machine type from default to 8 CPUs

### DIFF
--- a/ci/cloudbuild.yaml
+++ b/ci/cloudbuild.yaml
@@ -1,5 +1,8 @@
 steps:
 
+# This step will fetch the entire git history so tags can be accessed during
+# the build. If this negatively impacts build performance, please modify it so
+# only commits up to the latest tag is fetched.
 - name: 'gcr.io/cloud-builders/git'
   id: 'Fetch tags from git'
   args: [fetch, --unshallow]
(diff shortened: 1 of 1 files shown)


ci/cloudbuild.yaml
```diff
@@ -13,6 +13,10 @@ steps:
   id: 'Create version tags'
   args: ['python3', 'ci/create-version-tags.py']
 
+- name: 'gcr.io/cloud-builders/docker'
+  entrypoint: '/bin/bash'
+  args: ['-c', 'docker pull gcr.io/$PROJECT_ID/ts-bridge:latest || exit 0']
```
Temikus, line 18: This would hide any actual pull errors though. What do you want to do here? I may be able to offer something more selective here.

ci/cloudbuild.yaml
```diff
@@ -1,8 +1,8 @@
 steps:
 
-- name: gcr.io/cloud-builders/git
+- name: 'gcr.io/cloud-builders/git'
   id: 'Fetch tags from git'
-  args: [fetch, --depth=100, --tags]
+  args: [fetch, --unshallow]
```
Temikus, line 5: IIRC this will fetch all the commits from the beginning of time. Correct? Is this something we want?

Temikus (changes requested):
LGTM overall, just address 2 nits

Temikus (approved):

Issue #61, opened 2020-12-09
Migrate off of Travis-CI


Temikus:
We're using Travis-CI.org, which is getting shut down December 2020: https://docs.travis-ci.com/user/migrate/open-source-repository-migration

As such, we'll need to migrate to githubActions.

Temikus:
Assigning to Sophia to take a look as she's been working on integrating GH Actions for building images.

PR #69, opened 2020-12-23
Travis migration

SoapHia:
Migrate Travis CI (.travis.yml) to GitHub Actions, so that we can continue to run Go tests after Travis-CI shuts down on 31st Dec.

The GitHub Action workflow added will be triggered when there is a:

- push to master
- pull request to master
- release created

An example of the workflow output can be found at: https://github.com/SoapHia/ts-bridge/runs/1597974835
You can also compare this with an existing Travis run at: https://travis-ci.org/github/google/ts-bridge/builds/751122865

This PR closes google/ts-bridge#61

### DIFF
--- a/.github/workflows/golang-tests.yml
+++ b/.github/workflows/golang-tests.yml
@@ -1,9 +1,18 @@
-name: go-test
+name: Run Golang Tests
 on:
+  # Trigger the workflow on push or pull request,
+  # but only for the main branch.
   push:
-    # Run for all commits on test branch
     branches:
-      - travis-migration
+      - master
+  pull_request:
+    types: [opened, reopened, edited, synchronised]
+    branches:
+      - master
+  # Also trigger when a new release has been created.
+  release:
+    types:
+      - created
 jobs:
   install-and-run:
     name: Install go and run tests
--- a/.github/workflows/go-tests.yml
+++ b/.github/workflows/go-tests.yml
@@ -1,7 +1,7 @@
 name: Run Go Tests
 on:
   # Trigger the workflow on push or pull request,
-  # but only for the main branch.
+  # but only for the master branch.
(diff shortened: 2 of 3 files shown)


jaqx0r:
looks like they generate the same results!  Nice work.

.github/workflows/golang-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+    env:
+      CLOUDSDK_CORE_DISABLE_PROMPTS: 1
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Set up Go 1.13.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.13.0'
```
jaqx0r, line 29: I'll assume this is to maintain equivalence with Travis.  Should we make this always use the latest version of Go?  If not now, then please leave a TODO.

jaqx0r (changes requested):

.github/workflows/golang-tests.yml
jaqx0r: https://github.com/actions/setup-go/pull/32 implies that we can make this say '^1.' and it'll do the right thing.

.github/workflows/golang-tests.yml
```diff
@@ -0,0 +1,55 @@
+name: Run Golang Tests
```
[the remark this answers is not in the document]
SoapHia: Done :)

.github/workflows/golang-tests.yml
SoapHia: I've changed it to test using the latest version and added a TODO for future extension to other versions of Go.

jaqx0r (approved):

PR #75, opened 2020-12-30
Add GoReleaser to build binaries and populate changelog

### LINKED ISSUE #74: Automate release with Github Actions
Currently, changelogs and release descriptions are written manually for every release. We would like to use GitHub Actions to do the following:

- [ ] create/update CHANGELOG.md
- [x] create releases containing changelog which summarises all changes since previous release
- [x] Build and release binaries for platforms were possible: linux 86_64, arm, macOS
- [x] Release source archives in `tar.gz` format

SoapHia:
Works on #74 using [GoReleaser](https://goreleaser.com/). The configuration for GoReleaser is specified in `.goreleaser.yml`. 

This GitHub Action will be triggered by a change in the new `.ts-bridge-version` file, which should be updated whenever a new release is ready.

The GitHub Action will automatically create a release tag (with an empty body) for the commit, then run GoReleaser to create a draft release with the same tag. The draft release will have its body pre-filled with a changelog, with its assets being go binaries and a source archive. It will appear as follows:
image

Please view the test GitHub Action here: https://github.com/SoapHia/ts-bridge/actions/runs/453670288


nerdinary (commented):
How are version tags determined or set?
Can they be linked to the release version that is tracked here: https://github.com/google/ts-bridge/blob/master/version/version.go#L18 ?

SoapHia:
> How are version tags determined or set?
> Can they be linked to the release version that is tracked here: https://github.com/google/ts-bridge/blob/master/version/version.go#L18 ?

The current work flow is as follows:
1. A contributor manually creates a new release tag with empty content (v0.0.0)
2. GoReleaser GitHub Action is triggered and run
3. Draft release is created with tag v0.0.0
4. By clicking publish on the draft release, the contents of the draft replaces the existing v0.0.0 release.

So the tag used by this GitHub Action will have the same tag as the one manually created in step 1 above.

nerdinary:
> We may be able to change the GitHub action to be triggered by changes in `version.go`. That way we can automatically create releases with the version specified in that file. Would that be preferred?

Yup, I think that would make more sense. Else we'd be tracking two different "versioning" systems. I see two possible approaches.

a) Add a check to see if the new release tag matches the version in 'version.go', and make some test fail if they don't match.

OR

b) Trigger the Github action based on version changes in version.go.

I think (b) would keep things simple, having to only update the version in one place. I'll let you pick whichever option makes more sense.

SoapHia:
> I think (b) would keep things simple, having to only update the version in one place. I'll let you pick whichever option makes more sense.

Following our discussion offline, I have done the following:

- created a new file to track the project version (`version/.ts-bridge-version`)
- modified the Github Action to be triggered on changes to above file
- made the GH Action automatically create a tag for the commit with the version change. 
- updated this PR description to match above changes.

TODO before merging: make `version.go` set its `Revision()` constant with the new file.

.github/workflows/release-go-binaries.yml
```diff
@@ -0,0 +1,54 @@
+name: Tag and Release Go Binaries
+
+on:
+  push:
+    # paths: ["version/.ts-bridge-version"]
```
nerdinary, line 5: is this a comment you're meant to remove?

nerdinary (approved):

.github/workflows/release-go-binaries.yml
SoapHia: Oops yes thanks for picking that up

PR #80, opened 2021-01-08
Refactored of external dependencies.

WizardOfFaraz:
External dependencies are now created in main so we can fail fast if any external connectivity is not working.

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -139,7 +139,7 @@ func main() {
 
 	metrics, err := CreateMetrics(context.Background(), config)
 	if err != nil {
-		log.Fatalf("failed creating metric dependencies: %v", err)
+		log.Fatalf("failed initializing adaptor/collector dependencies: %v", err)
 	}
 	defer cleanup(metrics)
 
--- a/app/main.go
+++ b/app/main.go
@@ -137,13 +137,13 @@ func main() {
 		SyncPeriod:               *syncPeriod,
 	})
 
-	updateMetrics, err := Dial(context.Background(), config)
+	metrics, err := CreateMetrics(context.Background(), config)
 	if err != nil {
 		log.Fatalf("failed dialing dependencies: %v", err)
 	}
-	defer cleanup(updateMetrics)
+	defer cleanup(metrics)
 
-	h := web.NewHandler(config, updateMetrics)
+	h := web.NewHandler(config, metrics)
 	http.HandleFunc("/", h.Index)
 	http.HandleFunc("/sync", h.Sync)
 	http.HandleFunc("/cleanup", h.Cleanup)
@@ -160,7 +160,7 @@ func main() {
 	if !env.IsAppEngine() {
 		log.Debug("Running outside of appengine, starting up a sync loop...")
 		ctx, cancel := context.WithCancel(context.Background())
-		go syncLoop(ctx, cancel, config, updateMetrics)
+		go syncLoop(ctx, cancel, config, metrics)
 	}
 
 	// Build a connection string, e.g. ":8080"
@@ -172,7 +172,7 @@ func main() {
 
 }
 
-func syncLoop(ctx context.Context, cancel context.CancelFunc, config *tsbridge.Config, updateMetrics *tsbridge.UpdateMetrics) {
+func syncLoop(ctx context.Context, cancel context.CancelFunc, config *tsbridge.Config, metrics *tsbridge.Metrics) {
 	defer cancel()
 	for {
 		select {
@@ -180,7 +180,7 @@ func syncLoop(ctx context.Context, cancel context.CancelFunc, config *tsbridge.C
 			log.Debugf("Goroutines: %v", runtime.NumGoroutine())
 			ctx, cancel := context.WithTimeout(ctx, config.Options.UpdateTimeout)
 			log.WithContext(ctx).Debugf("Running sync...")
-			if err := tasks.Sync(ctx, config, updateMetrics); err != nil {
+			if err := tasks.Sync(ctx, config, metrics); err != nil {
 				log.WithContext(ctx).Errorf("error running sync() routine: %v", err)
 			}
 			cancel()
@@ -190,8 +190,8 @@ func syncLoop(ctx context.Context, cancel context.CancelFunc, config *tsbridge.C
 	}
 }
 
-// Dial connections to all external dependencies
-func Dial(ctx context.Context, config *tsbridge.Config) (*tsbridge.UpdateMetrics, error) {
+// CreateMetrics returns a Metrics struct containing external dependencies.
+func CreateMetrics(ctx context.Context, config *tsbridge.Config) (*tsbridge.Metrics, error) {
 	sd, err := stackdriver.NewAdapter(ctx, config.Options.SDLookBackInterval)
 	if err != nil {
 		return nil, fmt.Errorf("unable to initialize stackdriver adapter: %v", err)
@@ -200,14 +200,12 @@ func Dial(ctx context.Context, config *tsbridge.Config) (*tsbridge.UpdateMetrics
 	if err != nil {
 		return nil, fmt.Errorf("unable to initialize stats collector: %v", err)
 	}
-	return tsbridge.NewUpdateMetrics(ctx, sd, sc), nil
+	return tsbridge.New(ctx, sd, sc), nil
 }
 
-func cleanup(updateMetrics *tsbridge.UpdateMetrics) {
-	defer updateMetrics.StatsCollector.Close()
-	var err error
-	err = updateMetrics.SDClient.Close()
-	if err != nil {
+func cleanup(metrics *tsbridge.Metrics) {
+	defer metrics.StatsCollector.Close()
+	if err := metrics.SDClient.Close(); err != nil {
 		log.Fatalf("Could not close Stackdriver client: %v", err)
 	}
 }
--- a/app/main.go
+++ b/app/main.go
@@ -137,16 +137,17 @@ func main() {
 		SyncPeriod:               *syncPeriod,
 	})
 
-	h := web.NewHandler(config)
+	updateMetrics, err := Dial(context.Background(), config)
+	if err != nil {
+		log.Fatalf("failed dialing dependencies: %v", err)
+	}
+	defer cleanup(updateMetrics)
 
+	h := web.NewHandler(config, updateMetrics)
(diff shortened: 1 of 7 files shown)


tasks/tasks.go
```diff
@@ -55,28 +51,7 @@ func Sync(ctx context.Context, config *tsbridge.Config) error {
...
-		}
-	})
-
-	// Process errors from Once.Do blocks since we cannot return in those
-	if len(errSync) > 0 {
-		return fmt.Errorf("errors occured during sync init: %v", errSync)
-	}
-
-	if errs := tsbridge.UpdateAllMetrics(ctx, metrics, sdClient, config.Options.UpdateParallelism, statsCollector); errs != nil {
+	if errs := tsbridge.UpdateAllMetrics(ctx, metrics, config.Dependencies.SDClient, config.Options.UpdateParallelism, config.Dependencies.StatsCollector); errs != nil {
```
nerdinary, line 54: Sync is called in a loop, and each loop we're passing the connection handlers back and forth as args.

It might be cleaner to have a struct that holds these handlers, so that in the future if we add more external handlers, they can just be added to the struct and not passed around as args for every loop.
 
Maybe wrap UpdateAllMetrics in a struct?

```
type UpdateMetrics struct {
 sdClient       *stackdriver.Adapter
 statsCollector *tsbridge.StatsCollector
}

func New(ctx,  *stackdriver.Adapter, *tsbridge.StatsCollector) *UpdateMetrics {
 return &UpdateMetrics {
    ...
 } 
}

func (u *UpdateMetrics) All(ctx, config.Options.UpdateParallelism) error {
 ...
}

```

WDYT?


(or even wrap it around config.ExternalDependencies)

Temikus:
From a quick casual look this doesn't seem to be failing fast if there are issues with Stackdriver though.

For example, if I do not set auth and run it will just loop:

```
Response: {
  "error": "invalid_grant",
  "error_description": "reauth related error (invalid_rapt)",
  "error_subtype": "invalid_rapt"
}, name: custom.googleapis.com/datadog/concourse_database_queries
ncourse_database_queries: failed to get latest timestamp: GetMetricDescriptor error: rpc error: code = Unauthenticated desc = transport: oauth2: cannot fetch token: 400 Bad Request
Response: {
  "error": "invalid_grant",
  "error_description": "reauth related error (invalid_rapt)",
  "error_subtype": "invalid_rapt"
}, name: custom.googleapis.com/datadog/concourse_database_queries
```

tsbridge/config.go
```diff
@@ -2,11 +2,14 @@ package tsbridge
...
 	"time"
+
+	"github.com/google/ts-bridge/stackdriver"
 )
 
 // Config holds all the ConfigOptions
 type Config struct {
-	Options ConfigOptions
+	Options      ConfigOptions
+	Dependencies ExternalDependencies
```
Temikus, line 12: Nit: `ExternalDependencies` is a bit confusing as it leads me to think this is somehow related to library load or something. Maybe go with Initializers or something?

WizardOfFaraz:
> From a quick casual look this doesn't seem to be failing fast if there are issues with Stackdriver though.
> 
> For example, if I do not set auth and run it will just loop:
> 
> ```
> Response: {
>   "error": "invalid_grant",
>   "error_description": "reauth related error (invalid_rapt)",
>   "error_subtype": "invalid_rapt"
> }, name: custom.googleapis.com/datadog/concourse_database_queries
> ncourse_database_queries: failed to get latest timestamp: GetMetricDescriptor error: rpc error: code = Unauthenticated desc = transport: oauth2: cannot fetch token: 400 Bad Request
> Response: {
>   "error": "invalid_grant",
>   "error_description": "reauth related error (invalid_rapt)",
>   "error_subtype": "invalid_rapt"
> }, name: custom.googleapis.com/datadog/concourse_database_queries
> ```

Could you show me you triggered this error?

app/main.go
```diff
@@ -143,6 +144,10 @@ func main() {
 	http.HandleFunc("/cleanup", h.Cleanup)
 	http.HandleFunc("/health", h.Health)
 
+	if err := CreateExternalConnections(context.Background(), config); err != nil {
```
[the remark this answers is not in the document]
WizardOfFaraz: I added a cleanup function that gets deferred.

app/main.go
```diff
@@ -184,6 +189,20 @@ func syncLoop(ctx context.Context, cancel context.CancelFunc, config *tsbridge.C
 	}
 }
 
+// CreateExternalConnections will establish connections to all external dependencies
+func CreateExternalConnections(ctx context.Context, config *tsbridge.Config) error {
```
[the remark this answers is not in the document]
WizardOfFaraz: Used Dial here.

tasks/tasks.go
WizardOfFaraz: I wrapped around UpdateAllMetrics, wdyt?

tsbridge/config.go
WizardOfFaraz: See recommendation from nerdinary, went with UpdateMetrics, wdyt?

Temikus:
@WizardOfFaraz Regarding triggering.

Just don't configure the auth or unset auth and:
```
go build -o ts-bridge ./app && ./ts-bridge --debug --storage-engine=boltdb --stats-sd-project=graphite-sandbox --update-parallelism=4 --sync-period=10s
```

tsbridge/metric.go
```diff
@@ -38,6 +40,20 @@ type Metric struct {
...
 }
 
+// UpdateMetrics contains all external metric dependencies
+type UpdateMetrics struct {
+	SDClient       StackdriverAdapter
+	StatsCollector *StatsCollector
+}
+
+// NewExternalMetrics returns a UpdateMetrics struct.
+func NewUpdateMetrics(ctx context.Context, sd *stackdriver.Adapter, sc *StatsCollector) *UpdateMetrics {
```
nerdinary, line 50: readability nit:

For a func that just returns a struct with passed in values, `func New(...)` is the preferred convention. 

`NewUpdateMetrics.UpdateMetrics` is the pointer, which is redundant in naming.

`New.UpdateMetrics` is a lot cleaner.

app/main.go
```diff
@@ -184,6 +190,28 @@ func syncLoop(ctx context.Context, cancel context.CancelFunc, config *tsbridge.C
...
+	sc, err := tsbridge.NewCollector(ctx, config.Options.SDInternalMetricsProject, config.Options.MonitoringBackends)
+	if err != nil {
+		return nil, fmt.Errorf("unable to initialize stats collector: %v", err)
+	}
+	return tsbridge.NewUpdateMetrics(ctx, sd, sc), nil
+}
+
+func cleanup(updateMetrics *tsbridge.UpdateMetrics) {
+	defer updateMetrics.StatsCollector.Close()
+	var err error
```
nerdinary, line 208:
```
if err := updateMetrics.SDClient.Close(); err != nil {
 log....
}
```

tsbridge/metric.go
```diff
@@ -38,6 +40,20 @@ type Metric struct {
 	Record    storage.MetricRecord
 }
 
+// UpdateMetrics contains all external metric dependencies
+type UpdateMetrics struct {
+	SDClient       StackdriverAdapter
```
nerdinary, line 45: [Apologies, I realize I suggested UpdateMetrics in a previous comment, but reading through it again, I've added another commend. Sorry for making you revise these again].

readability nit:

The name of the struct should reflect what it holds. `UpdateMetrics` is probably a name better suited for a func that does the updates.

Prefer something like:

```
type Metrics struct {
 SDClient       StackdriverAdapter
 StatsCollector *StatsCollector
}
```

Then your func that does the actual update could be along the lines of `func (m *Metrics) UpdateAll(.....`

app/main.go
```diff
@@ -136,8 +137,13 @@ func main() {
 		SyncPeriod:               *syncPeriod,
 	})
 
-	h := web.NewHandler(config)
+	updateMetrics, err := Dial(context.Background(), config)
```
nerdinary, line 140: readability structure:

the name here is misleading, Dial just returns a struct - doesn't actually update any metrics (yet).

See my comment in `metric.go` about struct naming.

tsbridge/config.go
WizardOfFaraz: Moved this out of config and went with Metrics struct name instead.

tsbridge/metric.go
WizardOfFaraz: I went with Metrics struct name here.

app/main.go
WizardOfFaraz: Changed to CreateMetrics().  wdyt?

tsbridge/metric.go
WizardOfFaraz: Changed to func `New(...)`.

WizardOfFaraz:
> @WizardOfFaraz Regarding triggering.
> 
> Just don't configure the auth or unset auth and:
> 
> ```
> go build -o ts-bridge ./app && ./ts-bridge --debug --storage-engine=boltdb --stats-sd-project=graphite-sandbox --update-parallelism=4 --sync-period=10s
> ```

I tried the same and it fails before syncLoop starts:

```
./ts-bridge --debug --storage-engine=boltdb --stats-sd-project=ts-bridge-issue68 --update-parallelism=1 --sync-period=10s --enable-status-page
DEBU[0000] Debug logging enabled...
FATA[0000] failed creating metric dependencies: unable to initialize stackdriver adapter: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```

Temikus:
@WizardOfFaraz I think it's a corner-case of how I do auth, so let's not use this particular PR for settiling this.

LGTM

Temikus (approved):

app/main.go
```diff
@@ -136,8 +137,13 @@ func main() {
 		SyncPeriod:               *syncPeriod,
 	})
 
-	h := web.NewHandler(config)
+	metrics, err := CreateMetrics(context.Background(), config)
+	if err != nil {
+		log.Fatalf("failed creating metric dependencies: %v", err)
```
Temikus, line 142: Nit: this might be a tad confusing since this method doesn't really make any operations on metrics themselves rather just initializes adaptor/collector for them

WizardOfFaraz: Updated the log msg here.

nerdinary (approved):

PR #81, opened 2021-01-12
Run Go unit tests in different versions

SoapHia:
Adding functionality to run Go tests for the following versions:

- latest (1.15.x)
- major release before latest (1.14.x)
- project version (from .go-version)
- prerelease version (1.16beta1)

In order to use the project version in a job's matrix, I moved parsing of go-version into a separate job which the test job depends on.

All matrix jobs will run even if another matrix job fails. This is so that tests for different versions can all run independently.

The prerelease version has `continue-on-error` set to true so that the workflow will show as SUCCESSFUL (green tick) even if tests for the prerelease fails. The failed tests will still be annotated in the workflow summary.

### DIFF
--- a/.github/workflows/go-tests.yml
+++ b/.github/workflows/go-tests.yml
@@ -14,24 +14,38 @@ on:
     types: [created, edited, published]
 
 jobs:
-  install-and-run:
-    name: Install go and run tests
+  project-go-version:
+    name: Export current go version
     runs-on: ubuntu-latest
-    env:
-      CLOUDSDK_CORE_DISABLE_PROMPTS: 1
+    outputs:
+      go-version: ${{ steps.go.outputs.go-version }}
     steps:
       - name: Check out the repo
         uses: actions/checkout@v2
 
       - name: Set Go version
-        run: echo "go-version=$(cat .go-version)" >> $GITHUB_ENV
+        id: go
+        run: echo "::set-output name=go-version::$(cat .go-version)"
 
-      - name: Set up Go according to .go-version
+
+  install-and-run:
+    name: Install go and run tests
+    needs: project-go-version
+    runs-on: ubuntu-latest
+    env:
+      CLOUDSDK_CORE_DISABLE_PROMPTS: 1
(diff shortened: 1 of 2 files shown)


.github/workflows/go-tests.yml
```diff
@@ -41,6 +62,8 @@ jobs:
           sudo apt-get install openjdk-8-jre-headless
 
       - name: Run tests
+        # Allow tests to fail pre-release Go versions
+        continue-on-error: ${{ matrix.prerelease }}
```
nerdinary, line 66: If a test does fail for pre-release, what happens next? Is it just logged?

Temikus (approved):

jaqx0r (commented):
Looks good except I don't understand what the project-go-version job is about.

.github/workflows/go-tests.yml
SoapHia: it's just logged in the workflow outputs like below:
Screen Shot 2021-01-12 at 4 45 41 pm

SoapHia:
> Looks good except I don't understand what the project-go-version job is about.

We want to always run the unit tests on the version that ts-bridge is on, which is tracked in the file `.go-version`. 

I wasn't able to run `cat .go-version` inside the matrix config, so I had to do it in a separate job. `project-go-version` extracts this version and puts it into an output named `go-version`, which is later used via `needs.project-go-version.outputs.go-version` on line 41.

It's a bit complicated so please let me know if there's an easier way of doing this.

jaqx0r (approved):

.github/workflows/go-tests.yml
nerdinary: It's not obvious from the output that the failure is for a pre-release version. Can that be annotated somewhere?

SoapHia: As discussed offline I've updated the labels so that pre-release versions are specified explicitly. e.g.
image

jaqx0r (approved):

PR #82, opened 2021-01-13
Fix typos and correct naming for Python files

SoapHia:
- Correct Python scripts names to use underscores instead of dashes
- Change `path` to `paths` so workflow for Python unit tests is only triggered for relevant file changes (https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths)


Temikus (approved):

.github/workflows/run-python-tests.yml
```diff
@@ -2,7 +2,7 @@ name: Run Python Unit Tests
 
 on:
   push:
-    path: "ci/parse_trivy_results.py"
+    paths: "ci/parse_trivy_results.py"
```
Temikus, line 5: Just check if it's indeed a string, not an array.

Temikus (approved):

PR #90, opened 2021-01-14
update CI description, rename workflows to be more descriptive

SoapHia:
See the new README here: https://github.com/SoapHia/ts-bridge/tree/update-readme/.github/workflows

An error in trigger config for `go-releaser.yml` was found and corrected.

### DIFF
--- a/.github/workflows/test-docker-image-build.yml
+++ b/.github/workflows/test-docker-image-build.yml
@@ -27,6 +27,6 @@ jobs:
         uses: aquasecurity/trivy-action@master
         with:
           image-ref: ${{ env.image-tag }}
-          format: 'table'
+          format: "table"
           ignore-unfixed: true
-          exit-code: '1' # Exit with 1 if vulnerabilities listed below are found.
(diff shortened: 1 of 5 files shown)


.github/workflows/test-docker-image-build.yml
```diff
@@ -27,6 +27,6 @@ jobs:
         uses: aquasecurity/trivy-action@master
         with:
           image-ref: ${{ env.image-tag }}
-          format: 'table'
+          format: "table"
           ignore-unfixed: true
-          exit-code: '1' # Exit with 1 if vulnerabilities listed below are found.
+          exit-code: "1" # Exit with 1 if vulnerabilities listed below are found.
```
nerdinary, line 32: Typically, exit codes are ints, not strings. 

Does github actions expect a string?

nerdinary (approved):

.github/workflows/test-docker-image-build.yml
SoapHia: This particular action expects a string according to https://github.com/aquasecurity/trivy-action#customizing.

PR #99, opened 2021-02-08
Add local profiler and refactor LoadStorageEngine to be outside of syncLoop

SoapHia:
This PR moves LoadStorageEngine outside of the Sync routine for a ~30% improvement in its runtime.

**Context:**
By running the pprof profiler, I've discovered that a third of the main sync routine's time is spent loading storage. (In the flame graph below, `Sync` uses 3.17% of root, where `LoadStorageEngine` accounts of 0.79%)
image

Since `LoadStorageEngine` was called inside `tasks.Sync`, the storage engine (boltDB/datastore) would be reloaded for every sync cycle. The storage engine for TS-Bridge can only be set via runtime flags. Therefore, it will not change during runtime, making this repeated loading of the storage engine unnecessary. 

To optimise this function, I refactored `LoadStorageEngine` to only be called once at the start of the program. After this refactor, the profiler output (see flame graph below) showed that the time spent on `tasks.Sync` had reduced from 3.17% to 2.16%, resulting in a ~30% improvement in efficiency.
image

**Caveat:**
There may be issues with parallelism but the only go routines we use are syncLoop and writing to StackDriver, so the storage engine should still be unaffected.

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/google/ts-bridge/web"
 
 	"github.com/google/ts-bridge/env"
+	"github.com/google/ts-bridge/storage"
 	"github.com/google/ts-bridge/tsbridge"
 	"github.com/google/ts-bridge/version"
 
@@ -153,7 +154,13 @@ func main() {
 	}
 	defer cleanup(metrics)
 
-	h := web.NewHandler(config, metrics)
+	store, err := tasks.LoadStorageEngine(context.Background(), config)
+	if err != nil {
+		log.Fatalf("failed to load storage engine: %v", err)
+	}
+	defer store.Close()
+
+	h := web.NewHandler(config, metrics, store)
 	http.HandleFunc("/", h.Index)
 	http.HandleFunc("/sync", h.Sync)
 	http.HandleFunc("/cleanup", h.Cleanup)
@@ -161,7 +168,7 @@ func main() {
 
 	// Run a cleanup on startup
 	log.Debugf("Performing startup cleanup...")
-	if err := tasks.Cleanup(context.Background(), config); err != nil {
+	if err := tasks.Cleanup(context.Background(), config, store); err != nil {
 		log.Fatalf("error running the Cleanup() routine: %v", err)
 	}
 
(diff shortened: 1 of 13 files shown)


app/main.go
```diff
@@ -123,6 +129,11 @@ func main() {
 		log.Fatalf("Invalid flags: %v", err)
 	}
 
+	if *enableLocalCPUProfiler {
+		log.Debugf("CPU Profiler enabled.")
+		defer profile.Start(profile.CPUProfile, profile.ProfilePath(".")).Stop()
```
nerdinary, line 134: My understanding of defer, is that the function call isn't actually executed till main() exits.

I might be wrong here - but would this mean the profiler will only start once main exits, and then stop right after?

web/handlers_test.go
```diff
@@ -1,16 +1,19 @@
...
 	"testing"
 
+	"github.com/google/ts-bridge/tasks"
 	"github.com/google/ts-bridge/tsbridge"
 )
 
 func TestHealthHandler(t *testing.T) {
 	config := tsbridge.NewConfig(&tsbridge.ConfigOptions{})
-	h := NewHandler(config, &tsbridge.Metrics{})
+	store, err := tasks.LoadStorageEngine(context.Background(), config)
```
nerdinary, line 15: You need to validate `err` here before proceeding.

app/main.go
SoapHia: The documentation for this package states that this is the correct usage: https://github.com/pkg/profile

I believe that the profile.Start() is called immediately but .Stop() is not executed til main() exits. The profile produced with this code records the correct runtime duration so it seems to work as expected.

web/handlers_test.go
```diff
@@ -1,34 +1,63 @@
...
+			}
+			h := NewHandler(config, &tsbridge.Metrics{}, store)
 
-	req, err := http.NewRequest("GET", "/health", nil)
-	if err != nil {
-		t.Fatal(err)
-	}
+			req, err := http.NewRequest("GET", "/health", nil)
+			if err != nil {
+				t.Fatalf("Error while sending request for /health: %v", err)
```
nerdinary, line 44: readability nit: error strings must not be capitalized.

nerdinary (approved):
LGTM, one minor comment.

Temikus (approved):

PR #103, opened 2021-02-11
Refactored metric config to conditionally update metric config

SoapHia:
This PR splits tasks.Sync to avoid unnecessary reloads of the `metrics.yaml` file.

**Context**:
TSB users may potentially update their `metrics.yaml` file during runtime, and such changes must be propagated to the imported metrics quickly. Previously, this was achieved by reloading the metric config from `metrics.yaml` every sync loop via `NewMetricConfig(...)`. 

As shown in pprof profiler flame graphs, NewMetricConfig accounts for a large part of the update routine. 
image

**Changes:**
This PR refactors the code to only reload metric configs when the file has been modified. This was done by maintaining a cache of the `metrics.yaml`'s FileInfo (obtained using os.Stats(), which works for symlinks) within the `MetricConfig` struct. A file is assumed to be modified if both its `.Size()` and `.ModTime()` is different from those previously cached in the binary. 

**Result:**
the `syncLoop()` function now takes 0.63% of the total profile compared to the previous ~3%.
image

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -148,19 +148,20 @@ func main() {
 		SyncPeriod:               *syncPeriod,
 	})
 
-	metrics, err := CreateMetrics(context.Background(), config)
+	ctx, cancel := context.WithCancel(context.Background())
+	metrics, err := CreateMetrics(ctx, config)
 	if err != nil {
 		log.Fatalf("failed initializing adaptor/collector dependencies: %v", err)
 	}
 	defer cleanup(metrics)
 
-	store, err := tasks.LoadStorageEngine(context.Background(), config)
+	store, err := tasks.LoadStorageEngine(ctx, config)
 	if err != nil {
 		log.Fatalf("failed to load storage engine: %v", err)
 	}
 	defer store.Close()
 
-	metricCfg, err := tsbridge.NewMetricConfig(context.Background(), config, store)
+	metricCfg, err := tsbridge.NewMetricConfig(ctx, config, store)
 	if err != nil {
 		log.Fatalf("failed to perform initial load of metric config: %v", err)
 	}
@@ -173,7 +174,7 @@ func main() {
 
 	// Run a cleanup on startup
 	log.Debugf("Performing startup cleanup...")
-	if err := tasks.Cleanup(context.Background(), config, store); err != nil {
+	if err := tasks.Cleanup(ctx, config, store); err != nil {
 		log.Fatalf("error running the Cleanup() routine: %v", err)
 	}
 
@@ -181,7 +182,6 @@ func main() {
 	// TODO(temikus): refactor this to run exactly every SyncPeriod and skip sync if one is already active
 	if !env.IsAppEngine() {
 		log.Debug("Running outside of appengine, starting up a sync loop...")
-		ctx, cancel := context.WithCancel(context.Background())
 		go syncLoop(ctx, cancel, config, metrics, metricCfg, store)
 	}
 
--- a/app/main.go
+++ b/app/main.go
@@ -160,7 +160,12 @@ func main() {
 	}
 	defer store.Close()
 
-	h := web.NewHandler(config, metrics, store)
+	metricCfg, err := tsbridge.NewMetricConfig(context.Background(), config, store)
+	if err != nil {
+		log.Fatalf("failed to perform initial load of metric config: %v", err)
+	}
+
+	h := web.NewHandler(config, metrics, metricCfg, store)
 	http.HandleFunc("/", h.Index)
 	http.HandleFunc("/sync", h.Sync)
 	http.HandleFunc("/cleanup", h.Cleanup)
@@ -177,7 +182,7 @@ func main() {
 	if !env.IsAppEngine() {
 		log.Debug("Running outside of appengine, starting up a sync loop...")
 		ctx, cancel := context.WithCancel(context.Background())
-		go syncLoop(ctx, cancel, config, metrics, store)
+		go syncLoop(ctx, cancel, config, metrics, metricCfg, store)
 	}
 
 	// Build a connection string, e.g. ":8080"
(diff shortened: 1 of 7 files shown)


app/main.go
```diff
@@ -160,7 +160,12 @@ func main() {
 	}
 	defer store.Close()
 
-	h := web.NewHandler(config, metrics, store)
+	metricCfg, err := tsbridge.NewMetricConfig(context.Background(), config, store)
```
nerdinary, line 163: We should be using a single context for the entire program, and passing that context around.

```
ctx, cancel := context.WithCancel(context.Background())
metricCfg, err := tsbridge.NewMetricConfig(ctx, config, store)
```

Could you also update the CreateMetrics() and the syncLoop() calls, and re-use the same context there?

tasks/tasks.go
```diff
@@ -38,12 +40,37 @@ func LoadStorageEngine(ctx context.Context, config *tsbridge.Config) (storage.Ma
 	}
 }
 
-// Sync updates all configured metrics.
-func Sync(ctx context.Context, config *tsbridge.Config, metrics *tsbridge.Metrics, store storage.Manager) error {
-	metricCfg, err := tsbridge.NewMetricConfig(ctx, config, store)
+// Checks if the metric config file has been updated, and updates the cached file stats if it has.
```
nerdinary, line 43: https://golang.org/doc/effective_go.html#commentary

Comments must start with the name of the func. Also `metricConfigFileUpdated` is a mouthful. Perhaps `func isCfgUpdated` might be better, since it returns true if it is?

(or pick a more suitable name)

tasks/tasks.go
```diff
@@ -38,12 +40,37 @@ func LoadStorageEngine(ctx context.Context, config *tsbridge.Config) (storage.Ma
...
+func metricConfigFileUpdated(ctx context.Context, config *tsbridge.Config, metricCfgFs *os.FileInfo) (bool, error) {
+	currMetricCfgFs, err := os.Stat(config.Options.Filename)
```
nerdinary, line 45: Readbility nit:
In Go, variable names should be short, proportional to the size of its scope.
examples:
 'f' is preferred to 'file'
'c' is preferred to 'count'
'count' is preferred to 'countUsers'
... and so on

The func here already declares that it is dealing with the config, so the name doesn't need to repeat that info.

`f, err := os.Stat(....)`

tasks/tasks.go
```diff
@@ -38,12 +40,37 @@ func LoadStorageEngine(ctx context.Context, config *tsbridge.Config) (storage.Ma
...
+	if err != nil {
+		return false, err
+	}
+	if (*metricCfgFs).ModTime() != currMetricCfgFs.ModTime() && (*metricCfgFs).Size() != currMetricCfgFs.Size() {
```
nerdinary, line 49: Is there an instance where the `Size()` would change, and the `ModTime()` wouldn't?

you could just do

`return (*metricCfgFs).ModTime() != currMetricCfgFs.ModTime()), nil`

tasks/tasks.go
```diff
@@ -38,12 +40,37 @@ func LoadStorageEngine(ctx context.Context, config *tsbridge.Config) (storage.Ma
 	}
 }
 
-// Sync updates all configured metrics.
-func Sync(ctx context.Context, config *tsbridge.Config, metrics *tsbridge.Metrics, store storage.Manager) error {
-	metricCfg, err := tsbridge.NewMetricConfig(ctx, config, store)
+// Checks if the metric config file has been updated, and updates the cached file stats if it has.
+func metricConfigFileUpdated(ctx context.Context, config *tsbridge.Config, metricCfgFs *os.FileInfo) (bool, error) {
```
nerdinary, line 44: Is the necessary to pass the full config struct, when all you're reading is the filename?

tasks/tasks.go
```diff
@@ -38,12 +40,37 @@ func LoadStorageEngine(ctx context.Context, config *tsbridge.Config) (storage.Ma
...
+		return false, err
+	}
+	if (*metricCfgFs).ModTime() != currMetricCfgFs.ModTime() && (*metricCfgFs).Size() != currMetricCfgFs.Size() {
+		return true, nil
+	}
+	return false, nil
+}
+
+// SyncMetricConfig ensures that metric config is always synced with the metric config file. This should only be called when !env.IsAppEngine().
+func SyncMetricConfig(ctx context.Context, config *tsbridge.Config, store storage.Manager, metricCfg *tsbridge.MetricConfig) error {
```
nerdinary, line 56: In all these funcs, you're passing the same values over. the config struct, metric config etc.

Would it be cleaner to just store this in a struct? (please evaluate and see what's the best option here)

```
type Sync struct {
 cfg *tsbridge.Config
 metricCfg *tsbridge.MetricConfig
}
```

Then you can just call something like:

```
func (s *Sync) Config(ctx context.Context, store store.Manager) error {
 update, err := isCfgChanged(ctx)
....
}
```

This way you avoid passing the vars / pointers around and makes it easier to read.

nerdinary (commented):
Just did an initial review - I think it might be worth doing a bit of a rewrite to avoid passing around the same information to each func.

tasks/tasks.go
SoapHia: I changed it to `isMetricCfgUpdated` since there is another struct called `tsbridge.Config`.

tasks/tasks.go
SoapHia: thanks for the tip! I've updated it to be `fs` for file stats.

tasks/tasks.go
SoapHia: Changed this to pass only the filename :)

tasks/tasks.go
SoapHia: As discussed offline, this would require a lot of refactoring across `tasks.go` and `handlers.go`. To keep this PR small, I created an issue leave this refactor for future work: https://github.com/google/ts-bridge/issues/104

nerdinary: Sounds good.

nerdinary (approved):

PR #105, opened 2021-02-15
Benchmarks for StackdriverData

SoapHia:
Adding benchmark tests used for benchmarking StackdriverData() in InfluxDB and Datadog packages.

StackdriverData is a function which imports time series from external monitoring systems into TS-Bridge.
Test TimeSeries data with an exponential amount of data points (10, 100, 1000 and 10000) were generated and used on the StackdriverData method on both Datadog and InfluxDB implementations.


datadog/metric_test.go
```diff
@@ -37,6 +40,7 @@ import (
 )
 
 func TestMain(m *testing.M) {
+	log.SetOutput(ioutil.Discard)
```
nerdinary, line 43: I know this is being set to avoid noisy outputs for the benchmark tests, but will it also drop log outputs if a test run fails?

datadog/metric_test.go
```diff
@@ -351,3 +355,24 @@ func TestCounterStartTime(t *testing.T) {
 		})
 	}
 }
+
+func benchmarkStackdriverData(filename string, b *testing.B) {
```
nerdinary, line 359: Benchmark tests should be exported.

`func BencmarkStackdriverData(.....`

influxdb/metric_test.go
```diff
@@ -653,3 +655,28 @@ func TestStackdriverDataTimeAggCumulativeResponse(t *testing.T) {
 	expectedDesc, expectedTS = mustUnmarshalTimeSeries(expectedDescRaw, expectedTSRaw...)
 	compareTimeSeries(t, desc, expectedDesc, ts, expectedTS)
 }
+
+func benchmarkStackdriverData(filename string, b *testing.B) {
```
nerdinary, line 659: Benchmark tests should be exported.

`func BencmarkStackdriverData(.....`

nerdinary (approved):

datadog/metric_test.go
SoapHia: This is a helper function which is called in actual benchmark tests (line 374 onwards) so it is not exported. Should I add a comment in the code to note this?

nerdinary: Ah ofcourse - sorry I didn't spot that. All good.

datadog/metric_test.go
SoapHia: Discussed offline: test logs should be unaffected by this because they are logged using a separate `testing` package. However, I have changed the log level to `WarnLevel` so that any logs at warn/error/fatal level remain visible when running benchmark tests.

Issue #110, opened 2021-02-19
1.16.x (just released) - test failure


Temikus:
So 1.16 _just_ released and it seems that something has changed to make a couple of tests fail only for `1.16.x` build, notably:
- TestUpdateAllMetrics
- TestUpdateAllMetricsErrors

See: https://github.com/google/ts-bridge/runs/1933024386

Assigning to Adi to take a look later (this is not very urgent). Sorry for not doing it myself, but 1.16 just released and this didn't fail in RC so I couldn't have known.

nerdinary:
fixed in fixed in https://github.com/google/ts-bridge/pull/182

PR #128, opened 2021-04-14
Update go generated files for go ver 1.1[3-5]x

nerdinary:
go generate now adds a "." at the end of comment strings. updating the generated files to prevent test failures

eg. https://github.com/google/ts-bridge/runs/2063426061

go.mod will produce a diff for go 1.16x as our module dependencies are pinned to ver 1.13. This will be fixed in https://github.com/google/ts-bridge/issues/110


rjerrems (approved):
LGTM

PR #192, opened 2022-03-01
http2 package is causing failures due to dependency changes with older go versions.

nerdinary:
Dependencies now frequently seem to be failing for older go versions as some underlying methods have changed.

e.g.

```
# golang.org/x/net/http2
Error: ../../../go/pkg/mod/golang.org/x/net@v0.0.0-20220127200216-cd36cc0744dd/http2/transport.go:417:45: undefined: os.ErrDeadlineExceeded
```

Instead of creating multiple "if else" statements based on version, I'm bumping the minimum required version.


nerdinary:
Updated the github actions file to remove checks for old go versions.

Still keeping checks for Go 1.16x as that's version n-1, incase we need to rollback to an older Go version.

jaqx0r (approved):

PR #292, opened 2023-04-19
Update gcloud used in Go CI tests

kynan:
gcloud 307 fails with `gcloud failed to load: module 'collections' has no attribute 'MutableMapping'`, which is an incompatibility with Python 3.10.6 currently shipping with the [ubuntu-latest image](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md).

### DIFF
--- a/.github/workflows/go-tests.yml
+++ b/.github/workflows/go-tests.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Set up gcloud datastore emulator
         # JRE is needed for the datastore emulator
         run: |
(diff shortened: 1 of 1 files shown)


nerdinary:
LGTM on the change, but you won't be able to merge this as the trivy scanner is flagging vulnerabilities in the `golang.org/x/net` dependency, and that will need to be manually pinned to a newer version

nerdinary (approved):

nerdinary:
Or maybe you'll still be able to merge it. Either way, the dependency will need to be updated.

kynan:
The reason for this PR is precisely so I can merge the various dependabot PRs that address these vulnerable dependencies :)

